PR #6, opened 2020-02-25
Linting, renaming, general cleanup

JosephMontoya-TRI:
A few housekeeping things in structure.py, mostly formatting.  Also deleting a few redundant methods and parameters in functions.

* old version of `RawCyclerRun.get_interpolated_diagnostic_cycles` is removed.
* Moved newer version of `diagnostic_interpolated`->`get_interpolated_diagnostic_cycles`.
* Renamed `diagnostic_summary`->`get_diagnostic_summary`.


JosephMontoya-TRI:
Okay so a few more questions:

The current `RawCyclerRun.get_interpolated_diagnostic_cycles` has a few quirks that I want to make sure are okay (at least for now).  
* The interpolation is done at fixed endpoints outside of the range of voltages included for the diagnostic cycles, so there are a bunch of `np.nan` values in the resultant dataframe for values outside of the diagnostic cycle range.  Is this okay?
* Is it okay to have the diagnostic cycle as the "final" cycle?  I marked this in the code with a TODO on line 334.
* The test I've got isn't passing right now (i. e. the 4th value in diagnostic_cycles_interpolated for the test file is ~2.63), but I'm not sure it should be, is there an analogue to the old method that we should be testing against?

beep/run_model.py
```diff
@@ -600,10 +599,11 @@ def main():
     try:
         args = docopt(__doc__)
         input_json = args['INPUT_JSON']
+        # TODO: resolve test file dir
```
JosephMontoya-TRI: So one more note here, the pip build is failing because the default models are in the TEST_FILE_DIR.  Generally, we should never import the TEST_FILE_DIR in any non-test module because we want to be able to separate functional from testing pieces of data and code.  I've moved the models into their own directory and updated the manifest/setup.py accordingly.  Let me know if this is okay and I'll resolve the #TODO.

chirranjeevigopal-TRI: Agreed, and thanks for cleaning that up. I am ok with the changes.

beep/tests/test_structure.py
```diff
@@ -243,7 +243,7 @@ def test_get_interpolated_diagnostic_cycles_maccor(self):
         diagnostic_cycles_interpolated = \
             cycler_run.get_interpolated_diagnostic_cycles(min_n_steps_diagnostic=3,
```
chirranjeevigopal-TRI: Since the deprecated version of `get_interpolated_diagnostic_cycles()` was removed, we should discard this test as well?

beep/tests/test_structure.py
```diff
@@ -197,7 +197,7 @@ def test_to_diagnostic_interpolation_cycler_run(self):
 
         v_range, resolution, nominal_capacity, full_fast_charge, diagnostic_available = \
             cycler_run.determine_structuring_parameters()
-        diag_interpolated = cycler_run.diagnostic_interpolated(diagnostic_available)
+        diag_interpolated = cycler_run.get_interpolated_diagnostic_cycles(diagnostic_available)
         diag_cycle = diag_interpolated[(diag_interpolated.cycle_type == 'rpt_0.2C')
```
chirranjeevigopal-TRI: Lmk if you plan to, or if you would me like to add an assert for the additional grouping by `step_index_counter` that I included in a recent PR.

JosephMontoya-TRI: There's a revised test in `test_get_interpolated_diagnostic_cycles` that asserts that the partitioned steps have the same length, which roughly reflects their separate treatment, but I think ultimately we'll need to shore up the functionality more generally to have a more robust test.

beep/tests/test_structure.py
JosephMontoya-TRI: I think we can keep it, it should still work with the revisions I'd imagine?

PR #7, opened 2020-02-29
Mat 1439 add charge step to cycles interpolated

patrickherring-TRI:
This PR addresses a request to add charge steps to the interpolated cycles.
- Column 'step_type' added to interpolated frame. Typed as 'category', and set as either 'charge' or 'discharge' for filtering processes
- Function for filtering charge step
- Split up function into step interpolation function for either charge or discharge and function that combines the charge and discharge portions.

Minor clean up of big file tests is also included here.


beep/structure.py
```diff
@@ -132,29 +132,37 @@ def from_file(cls, path, validate=False):
         else:
             raise ValueError("{} does not match any known file pattern".format(path))
 
-    def get_interpolated_discharge_cycles(self, v_range=None, resolution=1000):
+    def interpolated_cycles(self, v_range, resolution, step_type='discharge'):
```
JosephMontoya-TRI: Is there a more descriptive name for this method that has an active verb and distinguishes it more from `get_interpolated_cycles`?  Maybe `get_interpolated_steps`?

beep/structure.py
```diff
@@ -132,29 +132,37 @@ def from_file(cls, path, validate=False):
...
             pandas.DataFrame: DataFrame corresponding to interpolated values.
         """
+        if step_type is 'discharge':
+            group = self.data.groupby(["cycle_index", "step_index"]).filter(
+                determine_whether_step_is_discharging).groupby("cycle_index")
+        elif step_type is 'charge':
+            group = self.data.groupby(["cycle_index", "step_index"]).filter(
+                determine_whether_step_is_charging).groupby("cycle_index")
+        else:
+            raise NotImplementedError
```
JosephMontoya-TRI: Can we make the error message more explicit?  Also might want to make it a ValueError, unless we're planning on implementing another step_type.

beep/structure.py
```diff
@@ -292,18 +318,22 @@ def get_diagnostic_summary(self, diagnostic_available):
...
+                                'cycle_index']
         diag_summary = diag_summary[diag_summary.index.isin(diag_cycles_at)]
 
         diag_summary['coulombic_efficiency'] = diag_summary['discharge_capacity'] \
                                                / diag_summary['charge_capacity']
-        diag_summary['diagnostic_type'] = list(itertools.chain.from_iterable(
-            [diagnostic_available['cycle_type'] for _ in starts_at]))
+
+        diag_summary['diagnostic_type'] = pd.Series(list(itertools.chain.from_iterable(
+            [diagnostic_available['cycle_type'] for _ in starts_at])))
```
JosephMontoya-TRI: You can create end-to-end joined copies of lists using multiplication e. g. [1, 2, 3] * 2 = [1, 2, 3, 1, 2, 3].

I think here it would look like list(diagnostic_available['cycle_type']) * len(starts_at).

Also not sure if the cast to pd.Series is necessary, but maybe I'm missing something.

beep/structure.py
```diff
@@ -1016,6 +1048,20 @@ def determine_whether_step_is_discharging(step_dataframe):
     return cap.diff(axis=0).mean(axis=0).diff().iloc[-1] > 0
 
 
+def determine_whether_step_is_charging(step_dataframe):
```
JosephMontoya-TRI: Where was this before?  Did it somehow get deleted?

beep/structure.py
chirranjeevigopal-TRI: Yeah, `interpolated_cycles` is also similar to `cycles_interpolated` attribute of the `processed_cycler_run` object.  I am fine with 'get_interpolated_steps` or `interpolate_within_step_type`

beep/structure.py
chirranjeevigopal-TRI: Previously, interpolation was only done on discharging steps, so we didn't need this method (to look for charging steps) until now.

beep/structure.py
```diff
@@ -165,6 +173,24 @@ def get_interpolated_discharge_cycles(self, v_range=None, resolution=1000):
 
         return result
 
+    def get_interpolated_cycles(self, v_range=None, resolution=1000):
+        """
```
chirranjeevigopal-TRI: Don't we want to filter out diagnostic cycles from being interpolated by this method?

beep/structure.py
patrickherring-TRI: Agreed, I think the methods should begin with `get` I will change that

beep/structure.py
patrickherring-TRI: Thanks for the tip! The casting to a series is necessary for cases when the diagnostic cycles have not completed and there is a length mis-match between the dataframe and the list. Generally, I would not do this, but in this case I think that the desired behavior is to take the length of the summary dataframe no matter what so truncating the list to the length of the dataframe is justified.

beep/structure.py
patrickherring-TRI: Agreed

beep/structure.py
patrickherring-TRI: I could filter to just the regular cycles, although it would add some complexity to the method. I will try and see if I can make it compact.

patrickherring-TRI: I think I found a relatively clean method for this

patrickherring-TRI:
@JosephMontoya-TRI I think I addressed all of your comments, thanks for those!

beep/structure.py
```diff
@@ -173,22 +177,39 @@ def interpolated_cycles(self, v_range, resolution, step_type='discharge'):
 
         return result
 
-    def get_interpolated_cycles(self, v_range=None, resolution=1000):
+    def get_interpolated_cycles(self, v_range=None, resolution=1000, diag_avail=None):
```
chirranjeevigopal-TRI: can we make the arg `diagnostic_available` for consistency with other methods?

beep/structure.py
chirranjeevigopal-TRI: Nice!

PR #14, opened 2020-03-18
Coverage

JosephMontoya-TRI:
This PR is for adding coverage reports to github status checks.

- Changes testing from nose (deprecated) to pytest
- Uses pytest-cov and coveralls for performing coverage checks and uploading checks to github
- Adds coverage badge to README


JosephMontoya-TRI:
Alright, looks like this is working now.

PR #18, opened 2020-03-20
Update the rest to increment by 12 min per protocol

patrickherring-TRI:
This PR is attempting to address a recurring problem where data acquisition stops when the HPPC cycles are running due to a data overload. Increasing the offset to 12 min per protocol should mean ~6 cells are going through the HPPC cycle at a time.


JosephMontoya-TRI:
Test is failing but otherwise LGTM.

PR #19, opened 2020-03-24
Increase the resolution of diagnostic interpolation from 500 to 1000

patrickherring-TRI:
This PR doubles the resolution of the diagnostic interpolation to address issues in fitting the response of the battery to pulses. This could potentially also create problems in the pipeline due to memory issues.

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -842,7 +842,7 @@ class ProcessedCyclerRun(MSONable):
 
     @classmethod
     def from_raw_cycler_run(cls, raw_cycler_run, v_range=None, resolution=1000,
-                            diagnostic_resolution=500, nominal_capacity=1.1,
+                            diagnostic_resolution=1000, nominal_capacity=1.1,
                             full_fast_charge=0.8, diagnostic_available=False):
         """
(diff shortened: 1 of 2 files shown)


beep/structure.py
```diff
@@ -842,7 +842,7 @@ def __init__(self, barcode, protocol, channel_id, summary,
 
     @classmethod
     def from_raw_cycler_run(cls, raw_cycler_run, v_range=None, resolution=1000,
-                            diagnostic_resolution=500, nominal_capacity=1.1,
+                            diagnostic_resolution=1000, nominal_capacity=1.1,
```
chirranjeevigopal-TRI: Since they wanted higher density for `hppc` cycles, why don't we use the finer mesh for just those instead of all diagnostic cycles?

patrickherring-TRI: Sure, I can change it for just the hppc cycles. It will be more brittle though, since it will then rely on being able to locate the hppc cycle

JosephMontoya-TRI:
Don't really know enough to know whether or not this is good from a science perspective, but looks fine for code 👍 .

PR #20, opened 2020-03-30
Appveyor windows

JosephMontoya-TRI:
A bit of polish left, but this PR should fix windows issues, mostly in pathing and how the tests pass json input from one script to the next.  Also adds an appveyor build.

Note that windows doesn't allow single-quotes to encapsulate string input on the command line, e. g. the `validate '{"mode": "events_off", ...}'` doesn't work.  I wrote a small function to format the json input for os-specific CLIs that formats the json on windows as `validate "{\"mode\": \"events_off\", ...}"`, which I think would also work on POSIX but is a little odd for usage messages, etc.


appveyor.yml
```diff
@@ -0,0 +1,27 @@
...
+    - PYTHON: "C:\\Miniconda3-x64"
+      PYTHON_VERSION: "3.7"
+      PYTHON_ARCH: "64"
+      platform: x64
+      STANDALONE: "FALSE"
+      CONDA_PY: "37"
+      CONDA_BUILD: "TRUE"
+      SPLIT_RUN: "FALSE"
+  PYPIPWD:
+    secure: cGsD88uzW0R371kFpYuczg==
```
moorepatrick: How does this security work?

JosephMontoya-TRI: Account-specific encryption, no decryption in PRs:
https://www.appveyor.com/docs/build-configuration/#secure-variables

JosephMontoya-TRI: I'm actually not a fan of using this build strategy anyway, I think the encryption is safe, but I'm going to remove it because I'd rather not do automated releases at this point.

Note that Travis does the same thing, except with the added bonus of encrypting the variable names as well.

patrickherring-TRI:
Thanks for figuring this out!

PR #21, opened 2020-04-03
Change method for interpolation of the HPPC cycle

patrickherring-TRI:
This PR changes the method for the interpolation of the HPPC cycle. Two problems exist in the current data; 1) the resolution of the voltage values is not high enough to allow for fitting of the kinetic parameters within a step 2) the data does not necessarily include the first and last voltage value for a step. This modification to the method attempts to address both of those problems while keeping the overall size of the data the same, by removing the rows that would normally have NaN values.

- Change the voltage interpolation for the hppc step to be on the voltage range in that iteration of the step.
- Set the resolution so that there is a default 0.5mV resolution for those steps
- Add cycle index as a column for the regular cycle summary (for reference development)

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -275,6 +275,7 @@ class RawCyclerRun(MSONable):
             reg_cycles_at = [i for i in self.data.cycle_index.unique()]
 
         summary = self.data.groupby("cycle_index").agg({
+            "cycle_index": "first",
             "discharge_capacity": "max",
             "charge_capacity": "max",
             "discharge_energy": "max",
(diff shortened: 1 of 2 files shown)


beep/structure.py
```diff
@@ -436,7 +438,7 @@ def get_interpolated_diagnostic_cycles(self, diagnostic_available,
         group = diag_data.groupby(["cycle_index", "step_index", "step_index_counter"])
         incl_columns = ["current", "charge_capacity", "discharge_capacity",
                         "charge_energy", "discharge_energy", "internal_resistance",
-                        "temperature", "datetime_seconds"]
+                        "temperature", "datetime_seconds", "test_time"]
```
chirranjeevigopal-TRI, line 441: Doesn't specifically matter for this PR, but how is the `test_time` field calculated? I recall we used to use it for computing `charge_time` and `time-temperature-integrated` features, and ran into some issues (negative values I think..).

patrickherring-TRI:
Yeah, the coverage dropped because there is a test that requires a large file and so is not run during the normal build process. I should look at a way of including that file since its becoming more critical to test the ability to parse the diagnostic data.

beep/structure.py
patrickherring-TRI: I think that the test_time for arbin can have some negative values right at the beginning of the test since the test_time is computed by the extraction script that I wrote. For the Maccor file this should never be the case since the test_time is output directly from the machine. In either case, we should really only be using this column for calculations requiring high precision, which should always take a delta in the time base and not the absolute value.

PR #25, opened 2020-04-08
MAT-1603 Reduce memory and reduce warnings

LindaHung-TRI:
Memory use is decreased, according to memory-profiler. Would appreciate confirmation that these new commands are equivalent to the previous version. Also added some .loc calls to get rid of pandas warning related to "Returning a view vs a copy" (https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#indexing-view-versus-copy).

[mem_profile_before.txt](https://github.com/TRI-AMDD/beep/files/4453146/mem_profile_before.txt)
[mem_profile_after.txt](https://github.com/TRI-AMDD/beep/files/4453148/mem_profile_after.txt)

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -151,22 +151,24 @@ class RawCyclerRun(MSONable):
         Returns:
             pandas.DataFrame: DataFrame corresponding to interpolated values.
         """
-        if step_type is 'discharge':
-            group = self.data.groupby(["cycle_index", "step_index"]).filter(
-                determine_whether_step_is_discharging).groupby("cycle_index")
-            group.apply(lambda g: g[g['cycle_index'].isin(reg_cycles)])
-        elif step_type is 'charge':
-            group = self.data.groupby(["cycle_index", "step_index"]).filter(
-                determine_whether_step_is_charging).groupby("cycle_index")
-            group.apply(lambda g: g[g['cycle_index'].isin(reg_cycles)])
-        else:
-            raise ValueError("{} is not a recognized step type")
-
         incl_columns = ["current", "charge_capacity", "discharge_capacity",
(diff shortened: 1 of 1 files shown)


beep/structure.py
```diff
@@ -152,21 +152,22 @@ def get_interpolated_steps(self, v_range, resolution, step_type='discharge', reg
...
+            step_filter = determine_whether_step_is_charging
         else:
             raise ValueError("{} is not a recognized step type")
-
         incl_columns = ["current", "charge_capacity", "discharge_capacity",
                         "internal_resistance", "temperature", "cycle_index"]
         all_dfs = []
-        for cycle_index, df in tqdm(group):
-            new_df = get_interpolated_data(df, "voltage", field_range=v_range,
+        cycle_indices = self.data.cycle_index.unique()
```
JosephMontoya-TRI, line 163: We might want to mention that we're purposefully not using the global groupby for the memory savings in a comment here, since we're doing something a bit non-idiomatic.  Something along the lines of "Use a cycle_index mask instead of a global groupby to save memory."

beep/structure.py
```diff
@@ -152,21 +152,22 @@ def get_interpolated_steps(self, v_range, resolution, step_type='discharge', reg
...
+        cycle_indices = [c for c in cycle_indices if c in reg_cycles]
+        cycle_indices.sort()
+        for cycle_index in tqdm(cycle_indices):
+            new_df = self.data.loc[self.data["cycle_index"] == cycle_index].groupby("step_index").filter(step_filter)
+            if new_df.size == 0:
+                continue
+            new_df = get_interpolated_data(new_df, "voltage", field_range=v_range,
```
chirranjeevigopal-TRI, line 170: I like this.

PR #26, opened 2020-04-10
MAT-1603b - Reduce peak memory usage and compute time

LindaHung-TRI:
- Change a dataframe.apply to a series.where, which significantly decreases memory usage and compute time
- Slight vectorization of loop contents, with small decrease in compute time
Peak memory usage should really be decreased this time...
mprof-150MiB-raw-beforepng
mprof-150MiB-raw-after

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -641,25 +641,31 @@ class RawCyclerRun(MSONable):
             state_type (str): charge or discharge.
 
         Returns:
-            list: summed quantities.
+            Series: summed quantities.
         """
         state_code = MACCOR_CONFIG["{}_state_code".format(state_type)]
-        quantity = data.apply(lambda row: row['_' + quantity] if row['_state'] == state_code else 0.0, axis=1)
-        earlier_quantity = 0.
-        earlier_cycle_index = data['cycle_index'][0]
-        new_step_flag = False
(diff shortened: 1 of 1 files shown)


beep/structure.py
```diff
@@ -641,25 +641,31 @@ def get_maccor_quantity_sum(data, quantity, state_type):
...
-                if cycle_index > earlier_cycle_index:
-                    earlier_quantity = 0.
-                    earlier_cycle_index = cycle_index
-                new_step_flag = False
-            quantity[i] += earlier_quantity
-            if (es >= MACCOR_CONFIG['end_step_code_min']) and (es <= MACCOR_CONFIG['end_step_code_max']):
-                new_step_flag = True
-                earlier_quantity += step_quantity
-        return quantity
+        quantity_agg = data['_' + quantity].where(data['_state'] == state_code, other=0.0, axis=0)
```
chirranjeevigopal-TRI, line 647: neat!

LindaHung-TRI:
Additional change to point out: the old code only reset the aggregated sum if the new cycle number was larger than the old cycle number. Now the code resets the aggregated sum if the cycle number is different (and can be smaller). In the case of a mid-file cycler restart, cycle number is reset to 0, and this change allows for the proper computation of the aggregated sums.

PR #30, opened 2020-04-15
MAT-1619

danielschweigert-TRI:
Changes to the featurization flow in order to lower the memory requirements during feature computation. Removing requirement of handling a pandas data frame with incoming data for feature computation and instead maintain data in a python dictionary.

### DIFF
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -97,10 +97,12 @@ class DegradationPredictor(MSONable):
             prediction_type (str): Type of regression - 'single' vs 'multi'.
             diagnostic_features (bool): whether to compute diagnostic features.
         """
-        processed_cycler_run = loadfn(path)
+
+        with open(path) as f:
+            processed_cycler_data = json.loads(f.read())
 
         if features_label == 'full_model':
-            return cls.init_full_model(processed_cycler_run, predict_only=predict_only,
+            return cls.init_full_model(processed_cycler_data, predict_only=predict_only,
                                        predicted_quantity=predicted_quantity,
                                        diagnostic_features=diagnostic_features,
                                        prediction_type=prediction_type)
@@ -108,7 +110,7 @@ class DegradationPredictor(MSONable):
             raise NotImplementedError
 
     @classmethod
-    def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycle=91,
+    def init_full_model(cls, processed_cycler_data, init_pred_cycle=10, mid_pred_cycle=91,
                         final_pred_cycle=100, predict_only=False, prediction_type='multi',
                         predicted_quantity="cycle", diagnostic_features=False):
         """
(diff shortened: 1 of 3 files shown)


beep/featurize.py
```diff
@@ -97,18 +97,20 @@ def from_processed_cycler_run_file(cls, path, features_label='full_model', predi
...
+            return cls.init_full_model(processed_cycler_data, predict_only=predict_only,
                                        predicted_quantity=predicted_quantity,
                                        diagnostic_features=diagnostic_features,
                                        prediction_type=prediction_type)
         else:
             raise NotImplementedError
 
     @classmethod
-    def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycle=91,
+    def init_full_model(cls, processed_cycler_data, init_pred_cycle=10, mid_pred_cycle=91,
```
chirranjeevigopal-TRI, line 113: Can you modify the doctstring to reflect the data type change?

beep/featurize.py
```diff
@@ -417,6 +421,87 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
     return json.dumps(output_data)
 
 
+def get_cycle_life(processed_cycler_data, n_cycles_cutoff=40, threshold=0.8):
```
chirranjeevigopal-TRI, line 424: For completeness sake, would it be possible to include unit tests for these? I suppose you can re-purpose the ones from `test_structure.py`, since the functionality shouldn't have changed.

beep/featurize.py
```diff
@@ -417,6 +421,87 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
...
+    """
+    Calculate cycle life for capacity loss below a certain threshold
+
+    Args:
```
chirranjeevigopal-TRI, line 428: Missing `processed_cycler_data` in the docstring

beep/featurize.py
```diff
@@ -417,6 +421,87 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
...
+    counter = 0
+    cycles = pd.DataFrame(np.zeros((1, len(threshold_list))))
+    for threshold in threshold_list:
+        cycles[counter] = get_cycle_life(processed_cycler_data, threshold=threshold)
+        counter = counter + 1
+    cycles.columns = np.core.defchararray.add("capacity_", threshold_list.astype(str))
+    return cycles
+
+
+def capacities_at_set_cycles(processed_cycler_data, cycle_min=200, cycle_max=1800, cycle_interval=200):
```
chirranjeevigopal-TRI, line 480: Missing `processed_cycler_data` in the docstring

beep/featurize.py
```diff
@@ -417,6 +421,87 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
...
+                      if processed_cycler_data['summary']['discharge_capacity'][x] < threshold * max_capacity])
+        cycle_life = ix_min
+    else:
+        # Some cells do not degrade below the threshold (low degradation rate)
+        cycle_life = len(processed_cycler_data['summary']['discharge_capacity']) + 1
+
+    return cycle_life
+
+
+def cycles_to_reach_set_capacities(processed_cycler_data, thresh_max_cap=0.98, thresh_min_cap=0.78, interval_cap=0.03):
```
chirranjeevigopal-TRI, line 458: Missing `processed_cycler_data` in the docstring

beep/featurize.py
```diff
@@ -128,44 +130,46 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
         Returns:
             beep.featurize.DegradationPredictor: DegradationPredictor corresponding to the ProcessedCyclerRun file.
         """
+        required_cycle_count = 100
+        if len(processed_cycler_data['summary']['discharge_capacity']) < required_cycle_count:
```
LindaHung-TRI, line 134: This can be replaced by final_pred_cycle?

beep/featurize.py
```diff
@@ -128,44 +130,46 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
...
         X = pd.DataFrame(np.zeros((1, 20)))
         labels = []
         # Discharge capacity, cycle 2 = Q(n=2)
-        X[0] = summary.discharge_capacity[1]
+        X[0] = processed_cycler_data['summary']['discharge_capacity'][1]
         labels.append("discharge_capacity_cycle_2")
 
         # Max discharge capacity - discharge capacity, cycle 2 = max_n(Q(n)) - Q(n=2)
-        X[1] = max(summary.discharge_capacity[np.arange(final_pred_cycle)] - summary.discharge_capacity[1])
+        X[1] = max([x - processed_cycler_data['summary']['discharge_capacity'][1] for x in processed_cycler_data['summary']['discharge_capacity'][:final_pred_cycle]])
```
LindaHung-TRI, line 151: This may be faster with something like
processed_cycler_data['summary']['discharge_capacity'][:final_pred_cycle].max() - processed_cycler_data['summary']['discharge_capacity'][1]

beep/featurize.py
```diff
@@ -128,44 +130,46 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
...
+        X[4] = np.nanmean(processed_cycler_data['summary']['charge_duration'][1:6])
         labels.append("charge_time_cycles_1:5")
 
         # Descriptors based on capacity loss between cycles 10 and 100.
-        Qd_final = interpolated_df.discharge_capacity[interpolated_df.cycle_index == ifinal]
-        Qd_10 = interpolated_df.discharge_capacity[interpolated_df.cycle_index == 9]
+        _ixf = [i for i, x in enumerate(processed_cycler_data['cycles_interpolated']['cycle_index']) if x == ifinal]
+        Qd_final = np.array(processed_cycler_data['cycles_interpolated']['discharge_capacity'])[_ixf]
+        _ix9 = [i for i, x in enumerate(processed_cycler_data['cycles_interpolated']['cycle_index']) if x == 9]
+        Qd_10 = np.array(processed_cycler_data['cycles_interpolated']['discharge_capacity'])[_ix9]
```
LindaHung-TRI, line 171: I'm a little confused what datatype _ixf and _ix9 are. They appear to be lists of integers, but then how do they work as indices in lines 168 and 170?

beep/featurize.py
```diff
@@ -185,46 +189,47 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
...
         labels.append("slope_discharge_capacity_cycle_number_91:100")
         labels.append("intercept_discharge_capacity_cycle_number_91:100")
 
-        IR_trend = summary.dc_internal_resistance[list(range(1, final_pred_cycle))]
-        if any(v == 0 for v in IR_trend):
-            IR_trend[IR_trend == 0] = np.nan
+        IR_trend = processed_cycler_data['summary']['dc_internal_resistance'][1:final_pred_cycle]
+        for i in range(len(IR_trend)):
+            if IR_trend[i] == 0:
+                IR_trend[i] = np.nan
```
LindaHung-TRI, line 217: This may be faster as IR_trend = np.where(IR_trend == 0, np.nan, IR_trend)

LindaHung-TRI:
I'm a little worried about the loss of readability and the fact that these modules are now less object oriented. For instance, if loadfn was not causing the memory spike, it would be nice to keep that call in to match how files are loaded and saved throughout beep.

For future reference, were you able to find which calls caused the memory spikes?

We also now have functions / methods in structure.py and featurize.py that have near-identical functionality. If this is an essential change, but we still want to preserve the methods within the class, we should have comments in these functions that there are similar ones in other files, so any future updates are applied in the right place(s).

JosephMontoya-TRI:
I strongly agree with Linda here, there should [only be one way](https://www.python.org/dev/peps/pep-0020/) to do this sort of thing, and if our objects are memory-inefficient we should have a better understanding of why, and preserve our existing APIs to the extent possible.

PR #32, opened 2020-04-20
Cast data types for all data frames in the structure

patrickherring-TRI:
This PR adds type casting for the structured data in an attempt to reduce memory required for subsequent processing.
- Schema for data types in each of the structured dataframes
- `astype` calls to cast each dataframe to those data types upon creation
- tests to ensure that data is typed correctly
- logic around reloading dataframes and casting the data types

Memory usage of `diagnostic_interpolated` and `cycles_interpolated` before and after data typing
Screen Shot 2020-04-20 at 12 47 32 AM

### DIFF
--- a/beep/conversion_schemas/structured_dtypes.yaml
+++ b/beep/conversion_schemas/structured_dtypes.yaml
@@ -1,0 +1,65 @@
+cycles_interpolated:
+  cycle_index: 'int32'
+  current: 'float32'
+  voltage: 'float32'
+  temperature: 'float32'
+  internal_resistance: 'float32'
+  charge_capacity: 'float64'
+  discharge_capacity: 'float64'
+  step_type: 'category'
+
+summary:
+  cycle_index: 'int32'
+  discharge_capacity: 'float64'
+  charge_capacity: 'float64'
+  discharge_energy: 'float64'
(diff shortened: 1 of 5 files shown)


beep/conversion_schemas/structured_dtypes.yaml
```diff
@@ -0,0 +1,65 @@
...
+  discharge_capacity: 'float64'
+  charge_capacity: 'float64'
+  discharge_energy: 'float64'
+  charge_energy: 'float64'
+  dc_internal_resistance: 'float32'
+  temperature_maximum: 'float32'
+  temperature_average: 'float32'
+  temperature_minimum: 'float32'
+  date_time_iso: 'object'
+  energy_efficiency: 'float64'
```
chirranjeevigopal-TRI, line 22: Trying to understand the rationale for 64 bit precision for efficiency -- is it because degradation could be very sensitive to small changes in energy or coulombic efficiency?

patrickherring-TRI: Yeah, I just thought that efficiency is something where the difference between 0.9999 and 0.999 is significant so adding more accuracy there might be helpful. If you think we can get away with float32, that would be better.

PR #34, opened 2020-04-21
Change logging to stdout to cleaner method

patrickherring-TRI:
This PR changes the way that logging to stdout is done in a stage environment to more closely mirror BEEP-Sync (which is cleaner and easier). It also adds some retry logic around setting up the logger so that credential errors don't immediately cause the container to crash.

### DIFF
--- a/beep/__init__.py
+++ b/beep/__init__.py
@@ -9,6 +9,8 @@ import sys
 from functools import partial
 import numpy as np
 import watchtower
+import time
+from botocore.exceptions import NoCredentialsError
 from tqdm import tqdm as _tqdm
 
 from .config import config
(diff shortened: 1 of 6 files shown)


beep/__init__.py
```diff
@@ -72,7 +75,16 @@
...
+        for _ in range(MAX_RETRIES):
+            try:
+                hdlr = watchtower.CloudWatchLogHandler(log_group='/stage/beep/services')
+            except NoCredentialsError:
+                time.sleep(10)
+                continue
+            else:
+                break
+        else:
+            raise NoCredentialsError
```
chirranjeevigopal-TRI, line 87: Just for my understanding, does the retry logic actually work a fraction of the time? If not, how does it help handle the credentials error?

patrickherring-TRI: When we tried this with the data-syncer, waiting 10 sec and retrying seemed to work almost 100% of the time. No idea why exactly, but my guess is the script is starting before the container can get the credentials from the role

PR #35, opened 2020-04-22
power waveform generation from velocity profiles

chirranjeevigopal-TRI:
This PR implements a method that generates a power waveform from an input velocity waveform. The model (for converting vel -> power) and default parameters are taken from literature.

Ref: Model-Based SEI Layer Growth and Capacity Fade Analysis for EV and PHEV Batteries and Drive Cycles, JECS, 161 (14) A2099-A2108 (2014)

Output power is computed both as absolute and scaled (-1 to 1) values.


beep/generate_protocol.py
```diff
@@ -940,6 +941,57 @@ def insert_rpt_cyclev2(self, proc_dict, start, nominal_capacity, diagnostic_para
...
+    # Power = Force * vel
+    # Force = Rate of change of momentum + Rolling frictional force + Aerodynamic drag force
+
+    # Method treats the time-series as is and does not interpolate on a uniform grid before computing gradient.
+    power = m * np.gradient(df.v, df.t) + rolling_resistance_coef * m * g * np.cos(
+        theta * np.pi / 180) + 0.5 * rho * drag_coef * frontal_area * (df.v - v_wind) ** 2
+
+    power = -power * df.v  # positive power = charge
+
+    return pd.DataFrame({'time(s)': df.t,
```
patrickherring-TRI, line 991: Just checking, the idea is that we then call another method to convert this profile to the maccor or arbin waveform format?

chirranjeevigopal-TRI: Correct. This would be a helper function. The conversion to .MWF file is very likely a more substantial effort, but there is a ticket in the backlog for it.

PR #38, opened 2020-05-03
Fix charge step interpolation for regular cycles

patrickherring-TRI:
This PR fixes some errors with the interpolation for regular cycles. Memory usage might be increased as a result, although this interpolation was already being done so the effect should be minimal. The processed object is significantly larger on disk (29MB -> 47MB). However, this change is necessary to complete some research on degradation features.
- Removes code that was deleting the charge step from regular cycles
- Changes charge interpolation to be based on capacity instead of voltage

### DIFF
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -388,8 +388,8 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
             processed_paths_list.append(path)
             processed_run_list.append(run_id)
             processed_result_list.append("incomplete")
-            processed_message_list.append({'comment':'Insufficient data for featurization',
-                                            'error': ''})
+            processed_message_list.append({'comment': 'Insufficient data for featurization',
+                                           'error': ''})
 
         else:
             processed_data = DegradationPredictor.from_processed_cycler_run_file(
@@ -409,7 +409,7 @@ def process_file_list_from_json(file_list_json, processed_dir='data-share/featur
             processed_run_list.append(run_id)
             processed_result_list.append("success")
             processed_message_list.append({'comment': '',
-                                            'error': ''})
+                                           'error': ''})
 
     output_data = {"file_list": processed_paths_list,
                    "run_list": processed_run_list,
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -137,7 +137,12 @@ class DegradationPredictor(MSONable):
         assert len(processed_cycler_run.summary) > final_pred_cycle, 'cycle count must exceed final_pred_cycle'
         cycles_to_average_over = 40  # For nominal capacity, use median discharge capacity of first n cycles
 
-        interpolated_df = processed_cycler_run.cycles_interpolated
+        if 'step_type' in processed_cycler_run.cycles_interpolated.columns:
+            interpolated_df = processed_cycler_run.cycles_interpolated[
+                processed_cycler_run.cycles_interpolated.step_type == 'discharge']
+        else:
+            interpolated_df = processed_cycler_run.cycles_interpolated
+
         X = pd.DataFrame(np.zeros((1, 20)))
         labels = []
         # Discharge capacity, cycle 2 = Q(n=2)
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -137,6 +137,7 @@ class DegradationPredictor(MSONable):
         assert len(processed_cycler_run.summary) > final_pred_cycle, 'cycle count must exceed final_pred_cycle'
         cycles_to_average_over = 40  # For nominal capacity, use median discharge capacity of first n cycles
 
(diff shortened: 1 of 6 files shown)


beep/featurize.py
```diff
@@ -137,7 +137,12 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
         assert len(processed_cycler_run.summary) > final_pred_cycle, 'cycle count must exceed final_pred_cycle'
         cycles_to_average_over = 40  # For nominal capacity, use median discharge capacity of first n cycles
```
chirranjeevigopal-TRI, line 138: Unrelated to the current PR, we were computing `nominal_capacity` for the fast charge project as median of discharge capacity over the first n=`cycles_to_average_over` cycles. This would lead to errors in the current projects (from looking at the profiles). Should we set nominal capacity based on a reference cycle, say 0 or 1?

patrickherring-TRI: Yeah, I think that what we should do is calculate the nominal capacity based on the feature set. So for the fast charge feature set, we might take the median of the maximum capacity over the first 10 cycles or something.

beep/structure.py
```diff
@@ -168,8 +169,13 @@ def get_interpolated_steps(self, v_range, resolution, step_type='discharge', reg
             new_df = self.data.loc[self.data["cycle_index"] == cycle_index].groupby("step_index").filter(step_filter)
             if new_df.size == 0:
                 continue
-            new_df = get_interpolated_data(new_df, "voltage", field_range=v_range,
-                                           columns=incl_columns, resolution=resolution)
+            if axis != 'voltage':
```
LindaHung-TRI, line 172: What are the other possibilites for "axis"? If it's only going to work for a few specific labels, this loop could be "if axis == 'voltage'", "elif axis == 'current'" (for example), and an "else" with NotImplementedError. Or alternately, to be more flexible, you could in line 161 add "axis" to the list, instead of "voltage" explicitly.

LindaHung-TRI: If only axis='voltage' works, that can also be asserted earlier in the script.

beep/featurize.py
```diff
@@ -137,7 +137,12 @@ def init_full_model(cls, processed_cycler_run, init_pred_cycle=10, mid_pred_cycl
...
 
-        interpolated_df = processed_cycler_run.cycles_interpolated
+        if 'step_type' in processed_cycler_run.cycles_interpolated.columns:
```
LindaHung-TRI, line 140: Curious to know when this "if" is triggered. Are there occasional cycles_interpolated that are unlabeled, but you know that they're discharge by default? Or are the cycles something entirely different?

patrickherring-TRI: Yeah, good point. I will add a comment. 
The functionality here is the same as it was, it's just that since we are adding charge steps to the interpolated data we need to filter the data for the discharge steps to get the same outcome as before. Benben should be adding some features that are dependent on the charge step in the regular cycles, so we will probably include a different feature set at that point.

beep/tests/test_structure.py
```diff
@@ -567,6 +579,23 @@ def test_from_raw_cycler_run_maccor(self):
...
+                (pcycler_run.cycles_interpolated.step_type == 'discharge')]
+            min_index_df = pcycler_run.cycles_interpolated[(pcycler_run.cycles_interpolated.cycle_index == min_index) &
+                                                           (pcycler_run.cycles_interpolated.step_type == 'discharge')]
+        else:
+            discharge_interpolated = pcycler_run.cycles_interpolated
+            min_index_df = pcycler_run.cycles_interpolated[(pcycler_run.cycles_interpolated.cycle_index == min_index)]
+        matches = discharge_interpolated.groupby("cycle_index").apply(
+            lambda x: np.allclose(x.voltage.values, min_index_df.voltage.values))
+        if not np.all(matches):
+            raise ValueError("cycles_interpolated are not uniform")
```
chirranjeevigopal-TRI, line 598: Can you help me understand what functionality is being tested by lines 586-598?. It seems to reproduce the `step_type` based decision logic that is also written into `structure`. Are you implicitly testing if this decision logic is being executed inside `structure`?

beep/structure.py
patrickherring-TRI: Changed this to have explicitly allowed axes for interpolation.

beep/tests/test_structure.py
patrickherring-TRI: It is testing if the interpolation has been applied uniformly, effectively moving that check from the class to the test. This PR removes that check in the initialization of the ProcessedCyclerRun class since it's not really relevant now. Also, given our need to reduce the memory and compute time for structuring, it seems like something that can be dropped.

PR #39, opened 2020-05-09
Mat 1656 refactor featurization

patrickherring-TRI:
This PR adds a framework for feature generation. There is a base class that outlines the required methods and provides the workflow for feature generation. The derived classes implement methods for the actual creation of the feature object.
- Base feature class
- Derived class `DeltaQFeatures`
- Derived class `DeltaQFeaturesSingle`
[Feature generation classes.pdf](https://github.com/TRI-AMDD/beep/files/4603164/Feature.generation.classes.pdf)

### DIFF
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -350,7 +350,7 @@ class DeltaQFeatures(BeepFeatures):
             processed_cycler_run (beep.structure.ProcessedCyclerRun): data from cycler run
 
         Returns:
-            pd.DataFrame: cycles at which capacity degradation exceeds thresholds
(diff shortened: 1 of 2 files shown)


beep/featurize.py
```diff
@@ -47,6 +47,357 @@
 s = {'service': 'DataAnalyzer'}
 
 
+class BeepFeatures(MSONable):
```
JosephMontoya-TRI, line 50: I would use the abc module here, e. g. make BeepFeatures an ABC or ABCMeta, use @abstractclassmethod and @abstractmethod for method handles.  Other than that looks good.

PR #41, opened 2020-05-11
Refactor featurization into class objects

patrickherring-TRI:
This PR adds a framework for feature generation. There is a base class that outlines the required methods and provides the workflow for feature generation. The derived classes implement methods for the actual creation of the feature object.

- Base feature class
- Derived class `DeltaQFeatures`
- Derived class `DeltaQFeaturesSingle`
[Feature generation classes.pdf](https://github.com/TRI-AMDD/beep/files/4611199/Feature.generation.classes.pdf)

### DIFF
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -572,11 +572,6 @@ class DegradationPredictor(MSONable):
         X[19] = summary.dc_internal_resistance[ifinal] - summary.dc_internal_resistance[1]
         labels.append("internal_resistance_difference_cycles_2:100")
 
-        if diagnostic_features:
-            X_diagnostic, labels_diagnostic = init_diagnostic_features(processed_cycler_run)
-            X = pd.concat([X, X_diagnostic], axis=1, sort=False)
-            labels = labels + labels_diagnostic
-
         X.columns = labels
         if predict_only:
             y = None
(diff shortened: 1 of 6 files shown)


patrickherring-TRI:
@LindaHung-TRI Do you mean having two base classes `Supervised` and `Unsupervised`? Right now there are several other assumptions that might prevent more general use of this feature class. Primarily we are assuming that `self.X` is a pandas dataframe, which might not be suitable for some features that are inherently higher dimensional objects.

LindaHung-TRI:
I'm not entirely sure what to call the new class, but my main thought is that the existence of features should be decoupled from how they're being used in the model. In this specific example of aggregation of cycling data into features, the "outcome" (number of cycles_to_reach_set_capacities) is also feature of the specific cycler run.  Note that when I say "feature", I don't mean an input to the model, but rather a distinguishing statistic / aggregation of the data.

One way to decouple features and how they're used if to have both a features class, plus a ModelXY (or ModelDataset, or some better name) class. The ModelXY class will be the one that takes features that have been generated, as well as lists of "X" names and "y" name(s), to prepare the X and y dataframes. ModelXY can further be subclassed to supervised and unsupervised, if desired, but as @patrickherring-TRI said, that's probably not too important for this project.

Especially if you're testing a variety of outcomes, it'll be convenient to have all relevant data aggregated in the featurization framework, and then you can swap out X and y as appropriate.

patrickherring-TRI:
@LindaHung-TRI I refactored the code slightly to accomodate a single feature object that can be either input or outcome depending on how we want to use it. Let me know if this is closer to what you were thinking? Happy to also jump on a video call tomorrow if that would be a better venue.

LindaHung-TRI:
Yes, looks good! And much more concise than my long comments above.

PR #42, opened 2020-05-14
MAT 1582 generate maccor waveform file from power waveforms

chirranjeevigopal-TRI:
This PR adds the ability to convert a variable power waveform (simulating driving profiles) into a maccor-readable waveform file (MWF). Output is path to a .mwf file which can directly be assigned to the appropriate step in the procedure dictionary (`proc_dict`). The  defaults for control mode, step termination criteria etc are read from a configuration file, if not provided by the user.

The ordering of columns and logic for assignment were inferred from a set of reference MWF files, with different sets of control parameters.


beep/generate_protocol.py
```diff
@@ -988,9 +989,76 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+    Returns:
+         path to the maccor waveform file generated
+    """
+
+    if mwf_config is None:
+        mwf_config = MACCOR_WAVEFORM_CONFIG
+
+    df['power'] = df['power'] / max(abs(df['power'])) * mwf_config['value_scale']
+
+    df['step_counter'] = df['power'].diff().fillna(0).ne(0).cumsum()
```
patrickherring-TRI, line 1018: Is this incrementing every time the power changes or flips sign?

beep/generate_protocol.py
```diff
@@ -988,9 +989,70 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+
+    df['control_mode'] = mwf_config['control_mode']
+
+    if mwf_config['control_mode'] == 'P':
+        df.loc[df['power'] == 0, 'control_mode'] = 'I'
+
+    df['value'] = np.round(abs(df['power']), 5)
+
+    mask = df['power'] <= 0
+    df = df.assign(**{'state': 'C',
```
patrickherring-TRI, line 1032: cool trick!

beep/generate_protocol.py
```diff
@@ -988,9 +989,70 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+                      'limit_mode': mwf_config['charge_limit_mode'],
+                      'limit_value': mwf_config['charge_limit_value'],
+                      'end_mode': mwf_config['charge_end_mode'],
+                      'operation':  mwf_config['charge_end_operation'],
+                      'end_mode_value': mwf_config['charge_end_mode_value']
+                      })
+
+    df.loc[mask, ['state', 'limit_mode', 'limit_value', 'end_mode', 'operation', 'end_mode_value']] =\
+        ['D', mwf_config['discharge_limit_mode'], mwf_config['discharge_limit_value'],
```
patrickherring-TRI, line 1041: Charge and discharge voltage limits might need to be changed from cell to cell, my guess is that we will pass them into the function as a portion of the parameters in the protocol. We can do that in a later PR?

patrickherring-TRI: Or actually, since you are already using the config, we can just send a different config when call this function.

beep/generate_protocol.py
chirranjeevigopal-TRI: Increments every time the power changes.

beep/generate_protocol.py
chirranjeevigopal-TRI: Yeah, there should be a method in the `ProcedureFile` class that calls this helper, and we will modify the default configuration as appropriate before doing so.

PR #43, opened 2020-05-15
Change the interpolation axis to be the max range across the run

patrickherring-TRI:
This PR changes the interpolation axis for regular cycles to be computed as the maximum range in the run instead of just interpolating over the range for each cycle. This is necessary in order to subtract of two capacity-voltage curves from different cycles within the same run, a quantity related to the delta Q feature.


patrickherring-TRI:
@chirranjeevigopal-TRI Yup, I added a test for that condition

PR #46, opened 2020-05-20
Add data types to arbin

patrickherring-TRI:
This PR adds data types to the Arbin raw data as an initial attempt to reduce the memory footprint.
- Added section to the arbin schema for data types
- Data types are applied immediately after loading the data into a dataframe

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -532,7 +532,7 @@ class RawCyclerRun(MSONable):
         for column, dtype in ARBIN_CONFIG['data_types'].items():
             if column in data:
                 if not data[column].isnull().values.any():
-                    data[column] = pd.Series(data[column], dtype=dtype)
+                    data[column] = data[column].astype(dtype)
 
         data.rename(ARBIN_CONFIG['data_columns'], axis='columns', inplace=True)
         metadata = pd.read_csv(metadata_path)
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -527,6 +527,7 @@ class RawCyclerRun(MSONable):
         metadata_path = path.replace(".csv", "_Metadata.csv")
         data = pd.read_csv(path)
         data.rename(str.lower, axis='columns', inplace=True)
+        data = data.astype(ARBIN_CONFIG['data_types'])
         data.rename(ARBIN_CONFIG['data_columns'], axis='columns', inplace=True)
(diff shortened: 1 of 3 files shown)


beep/tests/test_structure.py
```diff
@@ -592,6 +595,20 @@ def test_from_raw_cycler_run_arbin(self):
         self.assertEqual(pcycler_run.barcode, "EL151000429559")
         self.assertEqual(pcycler_run.protocol, r"2017-12-04_tests\20170630-4_65C_69per_6C.sdu")
 
+        all_summary = pcycler_run.summary
+        reg_dyptes = all_summary.dtypes.tolist()
```
LindaHung-TRI, line 599: Is reg_dyptes a typo of reg_dtypes, or does it have some other meaning?

beep/structure.py
```diff
@@ -527,6 +528,12 @@ def from_arbin_file(cls, path, validate=False):
         metadata_path = path.replace(".csv", "_Metadata.csv")
         data = pd.read_csv(path)
         data.rename(str.lower, axis='columns', inplace=True)
+
+        for column, dtype in ARBIN_CONFIG['data_types'].items():
+            if column in data:
+                if not data[column].isnull().values.any():
+                    data[column] = pd.Series(data[column], dtype=dtype)
```
LindaHung-TRI, line 535: This is skipping the type of an entire column if a single value is bad. You probably don't need the "if" to check for null values, and can just convert directly? Also, you can also change datatype using data[column].astype(dtype). Not sure about memory requirements of that vs your version.

beep/tests/test_structure.py
patrickherring-TRI: That's so weird. I must have had a typo in the first one and then used autocomplete for all of the rest. Should be fixed now

beep/structure.py
patrickherring-TRI: The problem is that NaN values are floats so if a column is missing some values and we set those to NaN, and then try to convert the data type to `int` there will be an error. The `if` statement is to address that situation and just leave the data column as its current type.
Thanks for spotting the `pd.Series`. I will switch over to the other method for type conversion, I think it's more consistent with the rest of the code.

LindaHung-TRI: I see -- I've only tried float64 to float32. Good to know about the issue with int.

PR #48, opened 2020-05-21
Mat 1724 add validation criteria for min cycle number

patrickherring-TRI:
This PR changes the validation criteria for the maccor files to require that the cycle number start below 2. This should make the 'p2' files invalid and address the errors during structuring these files.

### DIFF
--- a/beep/collate.py
+++ b/beep/collate.py
@@ -229,6 +229,10 @@ def process_files_json():
     # chdir into beep root
     pwd = os.getcwd()
     os.chdir(os.environ.get("BEEP_ROOT", "/"))
+    if not os.path.exists(SRC_DIR):
+        os.makedirs(SRC_DIR)
+    if not os.path.exists(DEST_DIR):
+        os.makedirs(DEST_DIR)
 
     meta_list = list(filter(lambda x: '_Metadata.csv' in x, os.listdir(SRC_DIR)))
     file_list = list(filter(lambda x: '.csv' in x if x not in meta_list else None, os.listdir(SRC_DIR)))
--- a/beep/collate.py
+++ b/beep/collate.py
@@ -9,7 +9,7 @@ Options:
     -h --help       Show this screen
     --version       Show version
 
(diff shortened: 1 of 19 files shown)


beep/tests/test_end_to_end.py
```diff
@@ -28,7 +27,7 @@ def setUp(self):
             kinesis = boto3.client('kinesis')
             response = kinesis.list_streams()
             self.events_mode = 'test'
-        except NoRegionError or NoCredentialsError as e:
+        except Exception as e:
```
JosephMontoya-TRI, line 30: Can we raise a warning here?  Since the only conceivable errors in the try block are those related to AWS, I'm fine to use a blank exception, but if someone is trying to use the AWS functionality and hasn't configured it correctly, we'd want them to know, rather than erroring silently.

beep/collate.py
```diff
@@ -229,6 +229,10 @@ def process_files_json():
     # chdir into beep root
     pwd = os.getcwd()
     os.chdir(os.environ.get("BEEP_ROOT", "/"))
```
JosephMontoya-TRI, line 231: If we're going to make directories for the user, we may want to change the default BEEP_ROOT to something we're more likely to have permissions in, like the repo itself.  We also may, as the issue notes, want to think about a more descriptive name, e. g. BEEP_PROCESSING_DIR, or something along those lines.

JosephMontoya-TRI:
Sorry, my comments somehow ended up here rather than on the previous PR.

beep/collate.py
patrickherring-TRI: Agreed. I will change the name to something a little more descriptive. I think we can set it so that certain environments default to the beep installation location and the `stage` and `prod` environments default to the root

beep/tests/test_end_to_end.py
patrickherring-TRI: Good idea, I am including a warning in there

PR #52, opened 2020-05-29
Add outcomes as features

chirranjeevigopal-TRI:
PR adds a new class for storing outcomes for to predict trajectory of degradation. 
- Validation criteria are inherited from the class DiagnosticCyclesFeatures
- Outcome is calculated as fractional amount of a quantity (most commonly discharge energy or capacity) remaining relative to the first cycle at every diagnostic cycle, for each diagnostic cycle type (e.g. 'hppc', 'reset' ). If there are 5 unique diagnostic_cycles, 2 degradation quantities of interest and 10 eligible sets of  diagnostic cycles, there will be 100 rows in X
- Currently, the threshold above which the outcomes are calculated is hard-coded. These will be made into class arguments in a follow-up PR.

Features and outcomes derived from diagnostic cycles are now part of the pipeline

### DIFF
--- a/beep/featurize.py
+++ b/beep/featurize.py
@@ -195,11 +195,14 @@ class DiagnosticCyclesFeatures(BeepFeatures):
             bool: True/False indication of ability to proceed with feature generation
         """
         conditions = []
-        conditions.append(hasattr(processed_cycler_run, 'diagnostic_summary'))
-        conditions.append(hasattr(processed_cycler_run, 'diagnostic_interpolated'))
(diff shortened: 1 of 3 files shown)


chirranjeevigopal-TRI:
Fair point on the class name. I do need to parameterize the threshold number of cycles beyond which to calc the outcomes (I also flagged that in the PR description). I would need to modify the base class and its methods to change that, and ensure it does not break other feature classes, so I was going to address that in a follow-up PR.

PR #55, opened 2020-05-30
Run tests in scratch dir

chirranjeevigopal-TRI:
Moved a few protocol and validation tests to run from within scratch directory, so as to avoid files output by unit tests from accidentally getting committed.


beep/generate_protocol.py
```diff
@@ -1055,7 +1055,7 @@ def generate_maccor_waveform_file(df, file_prefix, file_directory, mwf_config=No
     return MWF_file_path
 
 
-def generate_protocol_files_from_csv(csv_filename, output_directory, **kwargs):
+def generate_protocol_files_from_csv(csv_filename, output_directory=None, **kwargs):
```
patrickherring-TRI, line 1058: Just curious about the reason for making this a default? is it for testing purposes?

patrickherring-TRI: Ah, wait, I get it. This is fixing an inconsistency with the way the function is used later on

chirranjeevigopal-TRI: Yeah, validation_records.json was always being written into the procedure_template folder despite having an output_directory argument.

PR #56, opened 2020-06-02
WIP: Mat 1735 turn off events if secrets not accessible

chirranjeevigopal-TRI:
PR to fix an [issue](https://github.com/TRI-AMDD/beep/issues/54) relating to AWS credentialing. 
In the test class setup, I check if `secret_accessible(ENVIRONMENT)` returns True. If not, mode is set to `events_off`.  This should handle situations where a user has an AWS account, but not the right secrets for beep aws resources.

### DIFF
--- a/beep/tests/test_validate.py
+++ b/beep/tests/test_validate.py
@@ -13,8 +13,8 @@ import boto3
 from monty.tempfile import ScratchDir
 from beep.validate import ValidatorBeep, validate_file_list_from_json, \
     SimpleValidator
-from beep import S3_CACHE, VALIDATION_SCHEMA_DIR
-
+from beep import S3_CACHE, VALIDATION_SCHEMA_DIR, ENVIRONMENT
+from beep.utils.secrets_manager import secret_accessible
 TEST_DIR = os.path.dirname(__file__)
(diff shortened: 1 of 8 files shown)


beep/tests/test_validate.py
```diff
@@ -210,13 +218,17 @@ def test_validation_maccor(self):
...
+        if not secret_accessible(ENVIRONMENT):
             self.events_mode = "events_off"
+        else:
+            try:
+                kinesis = boto3.client('kinesis')
+                response = kinesis.list_streams()
+                self.events_mode = "test"
+            except Exception as e:
+                warnings.warn("Cloud resources not configured")
+                self.events_mode = "events_off"
```
danielschweigert-TRI, line 230: Since we use this same code snippet multiple times, I recommend creating a method for it which can be called instead.

chirranjeevigopal-TRI: Good thought -- pushing a new commit shortly.

JosephMontoya-TRI:
Sorry for being late, no revision-related comments.  As our testing functionality grows, we may want to create a base class that has some boilerplate testing functionality that we can reuse, see:
https://github.com/materialsproject/pymatgen/blob/master/pymatgen/util/testing.py

PR #63, opened 2020-06-19
Add test for procedure with waveform

patrickherring-TRI:
This PR creates a test for a procedure with a waveform step. This file needs to be tested with on the Maccor system to make sure that all of the settings are correct.
maccor_waveform_simple_test


beep/protocol/maccor.py
```diff
@@ -280,9 +280,21 @@ def insert_maccor_waveform_discharge(self, waveform_idx, waveform_filename):
 
         """
         steps = self['MaccorTestProcedure']['ProcSteps']['TestStep']
-        assert steps[waveform_idx]['StepType'] == "FastWave"
 
-        steps[waveform_idx]['StepValue'] = waveform_filename.split('.MWF')[0]
+        self.set('MaccorTestProcedure.ProcSteps.TestStep.{}.StepType'.format(waveform_idx), 'FastWave')
+        self.set('MaccorTestProcedure.ProcSteps.TestStep.{}.StepMode'.format(waveform_idx), '')
+        self.set('MaccorTestProcedure.ProcSteps.TestStep.{}.Ends'.format(waveform_idx), None)
```
chirranjeevigopal-TRI, line 286: Will `Ends` be assigned once we run a trial procedure with the waveform?

patrickherring-TRI: I could actually just leave the `Ends` alone and see if it works anyway.

patrickherring-TRI: Are the ending criteria in the waveform file already?

chirranjeevigopal-TRI: They are, but it is for each step within the mwf file. Not sure if maccor interprets those as the "Ends". I agree it makes sense to give this a shot and see if the run completes successfully.

patrickherring-TRI: Just to close the loop here. `Ends` and `Reports` have to be empty for a FastWave step.

PR #78, opened 2020-07-26
Mat 1842 debug maccor to arbin

patrickherring-TRI:
This PR contains fixes in order to get conversion from maccor to arbin working.


beep/protocol/maccor.py
```diff
@@ -1100,6 +1100,30 @@ def insert_rpt_cyclev2(self, start, nominal_capacity, diagnostic_params):
 
         return self
 
+    def set_skip_to_end_diagnostic(self, max_v, min_v, step_key='070'):
+        """
+        Helper function for setting the limits for that will skip to the end diagnostic
```
moorepatrick, line 1105: This sentence looks a little broken.

beep/protocol/maccor_to_arbin.py
```diff
@@ -40,26 +45,51 @@ def create_sdu(self, sdu_input_name, sdu_output_name):
...
+        safety_key = []
         for key in keys:
-            if "Schedule" in key:
-                del schedule["Schedule"][key]
+            if bool(re.match('Step[0-9]+', key)):
+                del schedule['Schedule'][key]
+            if bool(re.match('UserDefineSafety[0-9]+', key)):
+                safety_data.append(deepcopy(schedule['Schedule'][key]))
+                safety_key.append(key)
+                del schedule['Schedule'][key]
```
moorepatrick, line 65: Can this be turned into an if...else if? It will save the tiniest bit of time by not evaluating the second if we matched the first. Also, if we expect to match one more than the other, it should go first.

beep/protocol/maccor_to_arbin.py
```diff
@@ -128,6 +158,7 @@ def compile_to_arbin(
         35	Loop 3 (Cycling 30)	    65536	   13	  0			T1	    T2		00010 00000 00000 00000	01101
         36	Loop 4 (Cycling 100)	    0	   16	  0				    T4			                    10000
         37	Loop 5 (always true)	524288	   1	  0			T4	    CI		10000 00000 00000 00000	00001
+        15  Loop                       15      1      0         PVs             00000 00000 00000 01111 00001
```
moorepatrick, line 161: Having these columns all line up would look pretty slick.

patrickherring-TRI: Yup, thanks, I think the tab did not get converted correctly

beep/protocol/maccor_to_arbin.py
patrickherring-TRI: Yes, that's a good idea, thanks!

beep/protocol/maccor.py
patrickherring-TRI: got it! fixed

beep/protocol/maccor_to_arbin.py
```diff
@@ -265,9 +297,9 @@ def compile_to_arbin(
             elif "Do" in step_abs["StepType"]:
                 loop_counter = int(re.search(r"\d+", step_abs["StepType"]).group())
                 blank_step["m_szStepCtrlType"] = "Set Variable(s)"
-                blank_step["m_szCtrlValue1"] = str(2 ** (loop_counter + 15))
-                blank_step["m_szExtCtrlValue1"] = "0"
+                blank_step["m_szCtrlValue"] = str(2 ** (loop_counter + 15))
```
chirranjeevigopal-TRI, line 300: Are the hard-coded values for `blank_step["m_szCtrlValue"], blank_step["m_szCtrlValue1"]` project specific or project agnostic? Would it make sense to load them from a config file?

patrickherring-TRI: These are control values that perform specific actions like "increment cycle index" or "reset loop counter" so they are project agnostic. Once we get them working correctly they should not need to be changed.

patrickherring-TRI: These are control values that perform specific actions like "increment cycle index" or "reset loop counter" so they are project agnostic. Once we get them working correctly they should not need to be changed.

PR #80, opened 2020-07-29
Opening up the capacity limits for the xIris project

patrickherring-TRI:


beep/validation_schemas/schema-arbin-nmc-phev.yaml
```diff
@@ -1,6 +1,6 @@
 charge_capacity:
   schema:
-    max: 30.0
+    max: 100.0
```
chirranjeevigopal-TRI, line 3: Are these for the prius phev cells? What's the max capacity expected? 25 Ah?

patrickherring-TRI: Should be ~25Ah but an error in the schedule file seems to be producing some higher capacities

PR #81, opened 2020-08-01
[WIP] GitHub actions

JosephMontoya-TRI:
Draft of GH Actions for BEEP.  WIP, comments welcome.


JosephMontoya-TRI:
Also you can see the first report here: https://github.com/JosephMontoya-TRI/BEEP/runs/934339541?check_suite_focus=true

(pycodestyle is not happy)

.github/workflows/beep-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -e .[tests]
+    - name: pytest
+      env:
+        MPLBACKEND: "Agg"
+        BEEP_ENV: "dev"
+      run: |
+        pytest beep --color=yes --cov=camd --cov-report html:coverage_reports
```
patrickherring-TRI, line 36: Should this be: `pytest beep --color=yes --cov=beep --cov-report html:coverage_reports`?

JosephMontoya-TRI:
> LGTM! Thanks for adding this. How did you test the changes to Github actions?

The action runs on the branch wherever it's merged, so right now it's running on my fork.

moorepatrick:
@JosephMontoya-TRI Can you see if you can get it to run with https://github.com/nektos/act?

JosephMontoya-TRI:
> @JosephMontoya-TRI Can you see if you can get it to run with https://github.com/nektos/act?

@moorepatrick I just tried the tool, installing it is straightforward (with brew) but it seems to have some issues suggesting to me that it's not quite ready for primetime (e.g. parsing errors on standard parts of the python workflow).  However, using the linting (pycodestyle, flake8, pydocstyle) tools locally relatively straightforward and should be reproduced exactly according to our specifications with setup.cfg.  Testing locally is a little more complicated by local state, but should similarly just run via `pytest` with the command line.

JosephMontoya-TRI:
I've toned this PR down to just refactor testing with gh actions, and will issue a separate one to fix add linting and fix the issues related to linting.  Also, @chirranjeevigopal-TRI I deleted the `e` file in the root directory from the last PR, but want to double check that it's not necessary, so let me know if that's going to cause any problems.

.github/workflows/beep-lint.yml
```diff
@@ -0,0 +1,37 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.8]
```
moorepatrick: I doubt different python versions will change linting, but for symmetry, would keeping this at the max version used in testing and deployment make sense?

.github/workflows/beep-test.yml
```diff
@@ -0,0 +1,40 @@
+name: Testing - main
+
+on: [push, pull_request]
+
+jobs:
+  build:
```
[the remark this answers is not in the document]
JosephMontoya-TRI: Definitely.  Will change the other one to linting.  I noticed that they were both the same thing when I was using act, so hopefully this can disambiguate if we need to do postprocessing of tests, etc.

PR #84, opened 2020-08-10
[WIP] New release

JosephMontoya-TRI:
I've been experimenting a bit with options for how we might refactor releases.  I think it's still good that we have some human decision-making in the process (specifically on updating the changelog), so this first stab does the following:

* Refactors tasks.py->release.py to be run as a standalone script, rather than an invoke cli tool.
* release.py updates the version, updates the changelog with an in-terminal editor, tags the current commit with the version, and pushes the github release via the github api.

I'm a little torn on this, so could use some feedback on whether or not this is simpler, too brittle, etc.

Still todo:
* update admin.rst
* Implement pypi release workflow action
* docker build?


moorepatrick:
Is this PR still relevant?

JosephMontoya-TRI:
probably not worth automating right now, I think I'll close.  We might reopen if this becomes necessary.

PR #85, opened 2020-08-12
MAT-1870: Workflow Outputs

moorepatrick:
This PR creates additional output files on the local file system to support processing via Argo workflows.

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -1216,9 +1216,7 @@ class RawCyclerRun(MSONable):
                         )
                         diagnostic_starts_at.append(diag_cycle_num)
                     diagnostic_available = {
-                        "parameter_set": run_parameter["diagnostic_parameter_set"].iloc[
-                            0
-                        ],
+                        "parameter_set": run_parameter["diagnostic_parameter_set"].iloc[0],
                         "cycle_type": hppc_rpt,
                         "length": hppc_rpt_len,
                         "diagnostic_starts_at": diagnostic_starts_at,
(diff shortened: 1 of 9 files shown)


beep/structure.py
```diff
@@ -1183,18 +1198,27 @@ def determine_structuring_parameters(
...
+                                + hppc_rpt_len
+                            )
                             + 1
                             + run_parameter["diagnostic_start_cycle"].iloc[0]
                             + 1 * hppc_rpt_len
                         )
                         diagnostic_starts_at.append(diag_cycle_num)
                     diagnostic_available = {
-                        "parameter_set": run_parameter["diagnostic_parameter_set"].iloc[0],
+                        "parameter_set": run_parameter["diagnostic_parameter_set"].iloc[
```
patrickherring-TRI, line 1219: This is looks kind of strange... no problem with the code, but seems less readable

moorepatrick: I agree. As far as I can tell, it's an issue with how Black formats things.

I have not been impressed with how it splits long lines, and given that it's not configurable, I think we should consider using [yapf](https://github.com/google/yapf) instead.

beep/utils/events.py
```diff
@@ -354,6 +355,128 @@ def put_generate_event(self, output_data, generate_status):
         return response
 
 
+class WorkflowOutputs:
+    """
+    Attributes:
+        service (str): default: Which service is instantiating the class so that
+        we know who is sending events. Defaults to 'Testing'.
+        mode (str): Mode for events, test will only put test message. Defaults to 'run'.
+    """
```
chirranjeevigopal-TRI, line 364: Looks the docstring is carried over from class KinesisEvents and needs to be updated?

moorepatrick: 👍

PR #88, opened 2020-08-18
MAT-1932

willpowelson:
To convert Maccor protocol files to Biologic Protocol files we will need to be able to control what I'm calling "procedural steps" to sequences as opposed to the other type of Maccor step - "control steps".

Control steps are any of the following:
- Advancing Cycle Number
- Looping
- Do statements (beginning a cycle)

All other steps are procedural steps.

Procedural steps have a "control" portion to them, in that when a `<ReportEntry>` limit is reached they will go to some other arbitrary step. Since the mapping of steps to sequences is not 1-1 we'll need a more global context about the correspondence between step number from the input and sequence number in the output.

This behavior will be handled separately in a PR that's pending for the actual conversion.

### DIFF
--- a/beep/protocol/maccor_to_biologic_mb.py
+++ b/beep/protocol/maccor_to_biologic_mb.py
@@ -1,6 +1,11 @@
 # Copyright 2019 Toyota Research Institute. All rights reserved.
 """ Parsing and conversion of maccor procedure files to arbin schedule files"""
 
+import os
+from beep.protocol import PROTOCOL_SCHEMA_DIR
+from monty.serialization import loadfn
+from collections import OrderedDict
+
 
 class MaccorToBiologicMb:
     """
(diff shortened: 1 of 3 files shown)


beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -0,0 +1,301 @@
...
+    Collection of methods to convert maccor protocol files to biologic modulo bat protocol files
+    Differences in underlying hardware and software makes this impossible in some cases
+    or forces workarounds that may fail. Certain work-arounds may be attempted and are
+    documented under the conversion function.
+    """
+
+    def __init__(self):
+        self.blank_seq = {
+            # sequence number
+            "Ns": "0",
```
patrickherring-TRI, line 19: We might want to move this information into a .yaml file. See /beep/beep/protocol/protocol_schemas/arbin_schedule_schema.yaml

willpowelson: Ah, sure.

As a personal preference I prefer to keep as much context in a single file as possible to reduce cognitive load from switching between files but we should be consistant, will move this to a file named `biologic_modulo_bat_schema.yaml`

willpowelson:
@patrickherring-TRI 
Could I have some guidance on how test coverage works? Will I need test cases for every single error path to pass that check?

patrickherring-TRI:
@willpowelson All of the checks should pass, with the possible exception of coverage. In terms of test cases, I think the most important thing to capture in the tests are the intended functionality, not necessarily focusing on every method having a test (although that can be nice). If there are errors that you can foresee, having some kind of assert might help to catch problems in the future, but you don't need an exhaustive list of errors.

patrickherring-TRI:
@willpowelson Regarding the test coverage, I think the higher level philosophy is test driven development, writing the test as part of writing the code. But I think that's pretty similar to your normal workflow :)

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    def __init__(self):
+        BIOLOGIC_SCHEMA = loadfn(
+            os.path.join(PROTOCOL_SCHEMA_DIR, "biologic_mb_schema.yaml")
+        )
+        schema = OrderedDict(BIOLOGIC_SCHEMA)
+        self.blank_seq = schema["blank_seq"]
+
+    def __proc_step_to_seq(self, test_step, seq_num):
```
patrickherring-TRI, line 25: I don't think the double leading underscore is necessary. The convention I have seen is just one underscore

willpowelson: Ah, I didn't realize this! Can do!

PR #93, opened 2020-08-24
Create pull_request_template.md

moorepatrick:
This is an on-topic initial pull request template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,6 @@
+Fixes # .
+
+Changes proposed in this pull request:
+-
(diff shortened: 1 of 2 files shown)


HakyungKwon-TRI:
Looks like it doesn't like that the coverage has decreased?

moorepatrick:
Let me see if I can have pytest ignore those files.

moorepatrick:
I didn't look far enough into the coveralls report. None of the monitored files changed, so coverage should not have decreased. I think it's probably just a rounding/precision issue with Coveralls. I think it would be inappropriate to add additional changes to this PR to make that check pass.

JosephMontoya-TRI:
Yeah, something to do with AWS secrets exception getting uncovered, but definitely doesn't matter

PR #94, opened 2020-08-25
MAT-1934

danielschweigert-TRI:
Enables tests using big files which are imported from S3.


beep/tests/test_structure.py
```diff
@@ -26,10 +26,11 @@
 from monty.tempfile import ScratchDir
 from beep.utils import os_format
 from beep.utils.secrets_manager import event_setup
+from beep.utils.s3 import download_s3_object
 import matplotlib.pyplot as plt
 
-BIG_FILE_TESTS = os.environ.get("BEEP_BIG_TESTS", False)
-SKIP_MSG = "Tests requiring large files with diagnostic cycles are disabled, set BIG_FILE_TESTS to run full tests"
+BIG_FILE_TESTS = os.environ.get("BIG_FILE_TESTS", None) == "True"
```
patrickherring-TRI, line 32: Could we make the default false? Then I can set the environment variable in the build process. That way tests should always pass.

danielschweigert-TRI: `BIG_FILE_TESTS = os.environ.get("BIG_FILE_TESTS", None) == "True"` should default False if `BIG_FILE_TESTS` is not set.

patrickherring-TRI: Got it, thanks!

PR #95, opened 2020-08-26
MAT-1962 featurization failure

chirranjeevigopal-TRI:
Updated one of the helper functions for class HPPCRelaxationFeatures to handle errors caused due to occasional size mismatches between expected vs actual number of calculated SOC windows
In the fig attached, there should be only one relaxation curve around t=120000 and max 10 SOC windows. 

image

### DIFF
--- a/beep/features/featurizer_helpers.py
+++ b/beep/features/featurizer_helpers.py
@@ -836,9 +836,9 @@ def get_relaxation_features(processed_cycler_run, hppc_list=[0, 1], max_n_soc=10
 
         all_time_array = np.nan * np.ones((max_n_soc, 3))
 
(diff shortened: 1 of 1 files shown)


beep/features/featurizer_helpers.py
```diff
@@ -781,22 +781,25 @@ def get_relaxation_times(voltage_data, time_data, decay_percentage=[0.5, 0.8, 0.
     return np.array(time_array)
 
 
-def get_relaxation_features(processed_cycler_run, hppc_list=[0, 1]):
+def get_relaxation_features(processed_cycler_run, hppc_list=[0, 1], max_n_soc=10):
```
patrickherring-TRI, line 784: Wonder if it makes sense to ask Vivek to write a test for this method?

PR #98, opened 2020-09-09
MAT-1968: Output Results Files From generate_protocol

moorepatrick:
This PR adds outputting a results file when running `generate_protocol`.


patrickherring-TRI (approved):
LGTM! Thanks for adding this to the workflow!

PR #100, opened 2020-09-16
Add newline to end of init and setup

patrickherring-TRI:
This PR adds a newline to the end of the file when the release process is invoked so that the linting check passes.


JosephMontoya-TRI (approved):
Oh, that's what's going on.  Will look into this on other repos as well.

PR #102, opened 2020-09-23
MAT-1973 Biologic Modulo Bat file to Maccor Schedule File Converter (Part 1)

willpowelson:
This is the first half of a PR to convert Biologic `.mps` files with a single Modulo Bat Technique where cycle advancement is defined by loop iteration to a Maccor Schedule file.

Since the total changes would be over 1500 lines I'm looking for an initial review of the of the helper functions to convert a `sequence` from a Biologic file to a `step` in a Maccor file. If that passes muster, I plan to open a second PR where I effectively revert the simplification to this PR.

The full suite of tests/functionality can be found in commit `87845a6` for anyone in testing the conversion functionality in full.

@JosephMontoya-TRI
This is missing the main issues I was having with the linter, but I'm finding that the formatting from `black` is causing
```
missing whitespace after ','
```
error from `flake8` and `pycodestyle`, should I stop using `black` and revert to only dealing with errors from `flake8` and `pycodestyle`?


beep/protocol/biologic_mb_to_maccor.py
```diff
@@ -0,0 +1,404 @@
...
+        loop_to_seq = int(seq["ctrl_seq"])
+        loop_to_step = seq_num_by_step_num[loop_to_seq]
+
+        loop_end_entry["EndType"] = "Loop Cnt"
+        loop_end_entry["Oper"] = " = "
+        loop_end_entry["StepValue"] = "{}".format(loop_count)
+        loop_end_entry["Step"] = cls._stringify_step_num(loop_to_step)
+        assert len(loop_end_entry["Step"]) == 3
+
+        step["Ends"] = OrderedDict({"EndEntry": loop_end_entry,})
```
JosephMontoya-TRI, line 151: Ideally black would fix this, but here the issue is the trailing comma.  I'd just remove it manually for now.

JosephMontoya-TRI (approved):
I think we should consider using something like [pint](https://pint.readthedocs.io/en/stable/) in the long term for handling units and conversions, so we don't have to maintain that functionality, but for now this looks great!

beep/protocol/biologic_mb_to_maccor.py
JosephMontoya-TRI: See [this issue](https://github.com/psf/black/issues/1288).

willpowelson:
I agree for most of the codebase but I do actually think we want to avoid floating altogether when possible in converters like these. We're essentially transcribing a textual intent of a procedure in one format, to a (theoretically) equivalent textual representation in a different format.

Copying the actual text for a number and just slapping on the exponential term means we never interfere with a scientists intended specificity whereas with floating point someone might file a ticket complaining we transcribed `0.3` to `0.30000000000000004` :)

JosephMontoya-TRI:
> I do actually think we want to avoid floating altogether when possible in converters like these though.

I disagree here, I think it's a reasonable expectation that conversion between formats would be up to floating point precision, and we'll probably need to abide by some cycler-specific formatting rules for the outputs in the end anyway (for which we'll probably want to use float logic).  Also, exponential conversion only works if you're changing prefixes on a canonical base unit, can't convert kW to hp, for example, or even hr to s.  It's also just a lot to read and write tests for each switch case, which ultimately means more cognitive load for maintainers and new developers.  Just my two cents.

patrickherring-TRI:
I would agree with Joey here, most of the cyclers will expect a specific format for floating point, so I think we will have to write the formatter in any case.

willpowelson:
Hmmm, after some thought I think the long term solution is one of the following

Use `pint` (or something like it) to do conversions in decimal and then one of the following:

1. include a flag in the pubic API for conversion indicating whether or not truncate numbers to match cycler specificity constraints as determined by us. If set to false this would allow infinite specificity.
2. Dependency inject unit conversions and let users override our default ones

My thought is that, as a consumer of this library, I'd prefer any changes to significant figures to be a _deliberate_ choice I make, not something hidden in the software.

If I think no changes should have to be made to the significant figures specified in my Biologic file from my _particular_ Biologic cycler and lab software to my _particular_ Maccor cycler and lab software, I should be allowed to try that and watch it fail when loading into the lab software. Then _either_ I can trust TRI to do a good job on the conversion _or_ I can write a program that will adjust the significant figures (unit conversions are much easier than the structural conversions between these files).

Concretely, if a year from now a new Maccor model is released with higher specificity allowed for different metrics it should be possible for users to over-ride our conversion which are now _too_ specific for them without relying on us to make changes to the library or having to create a fork.

PR #105, opened 2020-09-23
MAT-1973 Biologic mb to Maccor converter (part 2)

willpowelson:
Here be dragons. 

This PR implements the structural transformations of a Biologic Modulo Bat protocol to an equivalent protocol in Maccor's Procedure format.

There are two main methods:
`convert` which intakes a file and writes a file
`biologic_mb_text_to_maccor_xml` which takes in the already loaded text from a file and returns the equivalent xml for Maccor.

There are a number of structural constraints that are imposed on all incoming files:
- Nested loops are forbidden, the way Maccor and Biologic implement nested loops is fundamentally incompatible as far as I can tell. From my reading a nested loop in Biologic will only loop over the inner loop during the _first_ loop of the outer loop. On subsequent iterations of the outer loop, the inner loop will run once. This is not representable on Maccor.
- In their documentation, Maccor forbids a GOTO into the the interior of a loop from outside that loop, you can only GOTO a "DO" step that begins a loop we also impose that constraint.
- In their documentation, Maccor forbids a GOTO that will move to a step before a completed loop, we also impose that constraint.

We provide helpful error messages in case of each of these violations.

### DIFF
--- a/beep/protocol/biologic_mb_to_maccor.py
+++ b/beep/protocol/biologic_mb_to_maccor.py
@@ -1,9 +1,13 @@
 # Copyright 2020 Toyota Research Institute. All rights reserved.
 """ Parsing and conversion of biologic Modulo Bat files to Maccor Procedure Files"""
 import os
+import re
 from copy import deepcopy
 from collections import OrderedDict
+import xmltodict
 from monty.serialization import loadfn
+from beep.protocol.biologic import Settings
+from beep.protocol.maccor import Procedure
 from beep.protocol import PROCEDURE_TEMPLATE_DIR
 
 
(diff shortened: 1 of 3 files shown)


beep/protocol/biologic_mb_to_maccor.py
```diff
@@ -398,3 +500,247 @@ def _create_step(
...
+        curr_loop_start = float("inf")
+        curr_loop_end = float("inf")
+
+        seqs_in_reverse_order = seqs.copy()
+        list.reverse(seqs_in_reverse_order)
+        for seq in seqs_in_reverse_order:
+            is_loop_seq = seq["ctrl_type"] == "Loop"
+            seq_num = int(seq["Ns"])
+            loop_to = int(seq["ctrl_seq"])
+            loop_is_empty = seq["ctrl_repeat"] == "0"
```
patrickherring-TRI, line 530: I seem to remember that if the repeat is 0 the cycler will just execute the next step, but it will also not increment the loop counter (so that the cycle index won't increment), is that right?

patrickherring-TRI (approved):
LGTM, I think we will need to validate the execution of the protocol once the cyclers are online.

beep/protocol/biologic_mb_to_maccor.py
willpowelson: That is correct, if a loop has the repeat set to zero it will be immediately passed over with no effect to cycle number. The only use for such a seq, as far as I'm aware, is in the construction of an "Advance Cycle" step.

patrickherring-TRI: Yeah, although there need to be two loop steps to create the equivalent of the "AdvCycle" step, right?

willpowelson: That is correct, this code allows for that specifically.  As an example of adding an "Advance Cycle" step in the EC-Lab software:
image
image
This adds a single advance cycle between seqs 0 and 3.
 
I've pushed a small naming change to alter this from "empty_loop" to "inactive_loop" and "non_empty_loop" to "active_loop", which I hope should be clearer.

willpowelson:
I forced push to avoid polluting git history with a lint fix.

PR #110, opened 2020-10-02
Fix error with length of list of peaks

patrickherring-TRI:
This PR addresses two errors that occur during featurization for a few files
- Uses length of the list of peaks instead of default value to address an out of index error
- Addresses an indexing error for the `get_fractional_quantity_remaining_nx` method


patrickherring-TRI:
@BruisVanVlijmen2020 Could you take a look at these changes as soon as you get a chance? If you approve the PR we should be able to go ahead and fix the errors in some of those feature objects.

chirranjeevigopal-TRI (approved):
LGTM!

BruisVanVlijmen2020 (approved):
It looks fine to me, but I'm not certain whether this is going to solve the indexing error that prompted these changes. 
In the current PR you're going off of len(peak_dQdVs). However, peak_dQdVs is an output to a function isolate_dQdV_peaks() which has max_nr_peaks as an input, so I would reckon max_nr_peaks and len(peak_dQdVs) are practically analogous.

patrickherring-TRI:
@BruisVanVlijmen2020 The method used will return _upto_ `max_nr_peak` but it can always return fewer peaks (at least that's my understanding from the scipy documentation). So  `max_nr_peak >= len(peak_dQdVs)`
https://github.com/TRI-AMDD/beep/blob/8a37edea16bbda526b093bf29e52e861b683a496/beep/features/featurizer_helpers.py#L97
The test `test_generate_dQdV_peak_fits` has a file where only 3 peaks are returned from the peak fitting.

PR #111, opened 2020-10-08
Make changes so that paused is the number of seconds 

patrickherring-TRI:
This PR changes the content of the paused metric in the summary to be the length of the gap between data points in seconds.

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -457,7 +457,7 @@ class RawCyclerRun(MSONable):
         self.data.drop(columns=["time_since_cycle_start"])
 
         # Determine if any of the cycles has been paused
-        summary["paused"] = self.data.groupby("cycle_index").apply(determine_paused)
+        summary["paused"] = self.data.groupby("cycle_index").apply(get_max_paused_over_threshold)
 
         summary = summary.astype(STRUCTURE_DTYPES["summary"])
 
@@ -530,7 +530,7 @@ class RawCyclerRun(MSONable):
             diag_summary["discharge_capacity"] / diag_summary["charge_capacity"]
         )
         diag_summary["paused"] = self.data.groupby("cycle_index").apply(
-            determine_paused
+            get_max_paused_over_threshold
         )
 
         diag_summary.reset_index(drop=True, inplace=True)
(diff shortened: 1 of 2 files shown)


beep/structure.py
```diff
@@ -2037,7 +2037,12 @@ def determine_paused(group, paused_threshold=3600):
         for t in date_time_obj
     ]
     date_time_float = pd.Series(date_time_float)
-    return int(date_time_float.diff().max() > paused_threshold)
+    if date_time_float.diff().max() > paused_threshold:
+        paused_secs = date_time_float.diff().max()
+    else:
+        paused_secs = 0
+
```
willpowelson, line 2044: I'm not sure I understand this. If the dataframe has say, 50 events I imagine as a researcher I want one of two things:

1. The sum of: each diff greater than `paused_threshold`
2. The sum of: each diff greater than `paused_threshold` minus paused threshold

(2) considers anything below pause threshold to be some artifact of the machine functioning, some noise if you will.

In our discussions I interpreted you as desiring (1).

This code would return only the maximum paused time that exceeded `paused_threshold`, is that what we want? I'd imagine for a dataframe like:
`[0.0, 3700.0, 7400.0]`

we'd return `7400.0` based on your verbal descriptions of this code but this would return `3700.0` as far as I can tell.

I think assigning `date_time_float.diff().max()` a name might help clear up this confusion for me.

willpowelson (changes requested):
>    Evaluate a raw cycling dataframe to determine if there is a pause in cycling
    Args:
        group (pd.DataFrame): cycling dataframe with date_time_iso column
        paused_threshold (int): gap in seconds to classify as a pause in cycling
    Returns:
        float: number of seconds that test was paused

I think this docstring is slightly inaccurate after these changes, at the very least we don't determine the if there was a pause, we measure the pause. I can't make a  specific recommendation on the updated wording because I'm slightly confused about the intent of the function, see my comment.

beep/structure.py
chirranjeevigopal-TRI, line 2044: I think I can see Will's source of confusion here. I suppose `paused_secs` could be renamed `max_pause_duration`, since you are only returning the `max` if there is >1 pause exceeding the threshold duration.

chirranjeevigopal-TRI (approved):
Variable naming aside, LGTM!

beep/structure.py
willpowelson: Ah, that would definitely explain it, thanks Chirru!

willpowelson:
Based on Chirru's comments it sounds like a small change to the naming should make this clearer:

rename the function `determine_paused` to something like `get_max_paused_over_threshold`
rename the variable `paused_secs` to `max_paused_over_threshold`

and update the docs accordingly.

patrickherring-TRI:
Got it, those variables should be clarified now.

willpowelson (approved):
LGTM!

PR #118, opened 2020-10-11
Create test and test file to resolve error

patrickherring-TRI:
This PR resolves a problem with the `get_v_diff` function used in featurization. For some files the change in the interpolation axis from point to point to smaller than the floating point precision used in `interp1d` This results in the interpolation failing due to a non-strict monotonic trend in the axis `ValueError: Expect x to be a 1-D sorted array_like.`
- Added a test file where this problem occurs.
- Changed the method to remove redundant points when they exist
- Added a test with a file that does cause the error and file that does not to ensure that values are not changed due to the alteration in the method.

### DIFF
--- a/beep/features/featurizer_helpers.py
+++ b/beep/features/featurizer_helpers.py
@@ -590,7 +590,7 @@ def get_V_I(df):
     return result
 
 
-def get_v_diff(processed_cycler_run, diag_pos, soc_window):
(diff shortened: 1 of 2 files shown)


JosephMontoya-TRI (approved):
LGTM.

chirranjeevigopal-TRI (approved):
LGTM... I am having a test fail for 'get_maccor_quantity_sum()' method for similar reasons of precision. :-)

PR #119, opened 2020-10-13
MAT-2042: Setup Coverage Config

moorepatrick:
This PR sets up pytest to ignore test files in the coverage report.

Note: Coverage has decreased and it appears that [`/beep/utils/secrets_manager.py:55`](https://coveralls.io/builds/34090576/source?filename=beep%2Futils%2Fsecrets_manager.py#L55) may be the cause. It looks like this line is covered in tests on the master branch, but not in PR branches.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = beep/tests/*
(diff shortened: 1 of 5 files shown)


patrickherring-TRI (approved):
LGTM!

PR #125, opened 2020-10-13
Biologic to maccor tests

patrickherring-TRI:
This PR just moves strings into a separate file to fit within the testing framework for the package.


patrickherring-TRI:
@chirranjeevigopal-TRI This is a fix for the failing test in the BEEP-EP repo

willpowelson (commented):
@patrickherring-TRI 
I'm not sure I understand these changes. This test file originally had

1. unit tests of individual functions
2. an integration test that intentionally did not include file decoding/encoding.
3. an integration test that did full file conversion

(2) was important for being able to quickly iterate on changes to the code, and this PR removes it. I originally had test a setup more like the one implemented here and found it frustrating to work with when I wanted to make changes to the `.mps` file with its LATIN-1 encoding. The tests using `biologic_mb_to_maccor_test_strs.py` was significantly more ergonomic.

You mentioned this will fix a failing test in BEEP-EP, could you expand on that? I'd prefer a solution that doesn't make iterating on this code more difficult if we can find one.

patrickherring-TRI:
@willpowelson The thing that is breaking is stemming from putting test data into a *.py file (a bit unusual) and requires that an init.py included in the test directory. If the idea is to have an "unencoded" version of the data, I guess that the best thing would be to put those variables directly in the test file. The other thing that I could try is just using utf-8 for that file, would that still serve the purpose of the test?

willpowelson:
It would, though I'm a bit worried about how difficult it is to document the reason that file would exist and was not encoded in LATIN-1, a `.py` file made it clear that the text was meant to be used as a simple Python string.

Is there no python convention for a file that simply stores and exports constants? All I really wanted was not have several hundred lines of test string constants in the test file...

patrickherring-TRI:
@willpowelson You can pickle a variable, but that's generally considered very bad practice for anything public since it allows arbitrary code execution... The main complication is that you need to store it as a string literal, not just a standard string?

willpowelson:
The issue was one of code readability. Originally I had the strings at the top of the test file, but found it made the file less comprehensible. In other languages I've worked in a file exporting constants was the way that problem was generally solved which was why I attempted it here.

We could move the string literals back into `test_biologic_mb_to_maccor` if you think that's acceptable. If not, placing them in UTF-8 text files while not ideal would suffice. If you place the strings into text files, I recommend the commit message mention the UTF-8 encoding.

patrickherring-TRI:
@willpowelson I experimented with a couple of different things and it looks like loading from a json file works fine (standard utf-8 encoding). Does that solution make sense?

willpowelson:
Works for me!

patrickherring-TRI:
@willpowelson Should be ready for re-review now

willpowelson (approved):
This will work, let's merge up!

PR #127, opened 2020-10-15
Add test for driving profiles at the console level

patrickherring-TRI:
This PR adds test coverage for validation and protocol generation.
- Re-enabled testing for cerberus based validation, could also remove this class since its not being actively used.
- Added console level testing for the newer portions of the protocol generator


chirranjeevigopal-TRI (approved):
LGTM!

PR #128, opened 2020-10-18
Change the method for ensuring monotonicity and add testing 

patrickherring-TRI:
This PR updates the `get_v_diff` method so that values are strictly monotonic with increase between values of `discharge_capacity` at least as big as the threshold.
- Threshold is computed as `(np.max(Q) - np.min(Q)) / 1e6`

### DIFF
--- a/beep/features/featurizer_helpers.py
+++ b/beep/features/featurizer_helpers.py
@@ -10,6 +10,7 @@ All methods are currently lumped into this script.
 import pandas as pd
 import numpy as np
 import matplotlib.pyplot as plt
+import os
 from scipy import signal
 from lmfit import models
 from scipy.interpolate import interp1d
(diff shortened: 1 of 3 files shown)


beep/features/featurizer_helpers.py
```diff
@@ -633,21 +633,27 @@ def get_v_diff(processed_cycler_run, diag_pos, soc_window, baseline_step_index=1
...
     V = chosen_1.voltage.values
     Q = chosen_1.discharge_capacity.values
 
-    #  Round values so that non-strictly monotonic values are removed
-    np.around(Q, decimals=6, out=Q)
-    np.around(V, decimals=6, out=V)
-    if not np.all(np.diff(Q) > 0):
-        index_of_repeated = np.where(np.abs(np.diff(Q)) == 0)[0]
+    #  Threshold between values so that non-strictly monotonic values are removed
+    d_capacity_min = (np.max(Q) - np.min(Q)) / 1e6
```
chirranjeevigopal-TRI, line 644: where does the 1e6 originate from? Something to do with the instrument resolution?

patrickherring-TRI: It's just a threshold for the minimum resolution, i.e the maximum number of data points that we should use is 1e6. This is somewhat arbitrary, but does link back to the overall precision of the instrument (most ADC's are 24 bit). The problem here is somewhat sloppy coding, and some under-documented features of interp1d (the interpolation axis for the function does not have the same precision required for all modes)

chirranjeevigopal-TRI (approved):
LGTM!

PR #130, opened 2020-10-20
Mat 1862 driving waveform structuring

chirranjeevigopal-TRI:
PR that allows processing maccor runs with waveform discharge.
- `get_maccor_quantity_sum()` updated to use waveform step specific quantities.
-  interpolation for regular cycles done on `test_time` if a cycle is detected to have a waveform step
-  expose interpolation axes in `from_raw_cycler_run`


beep/structure.py
```diff
@@ -230,6 +231,12 @@ def get_interpolated_steps(
             )
             if new_df.size == 0:
                 continue
+
+            # If cycle contains a waveform step, interpolate on test_time.
+            if self.data.loc[self.data["cycle_index"] == cycle_index].\
```
patrickherring-TRI, line 236: Having this buried in the structuring can lead to some unexpected behavior since it will silently switch over to time based interpolation if it decides there is a waveform step. Is there a way to make this decision logic more exposed?

beep/structure.py
```diff
@@ -1773,6 +1800,23 @@ def from_dict(cls, d):
         return cls(data, metadata)
 
 
+def determine_whether_step_is_waveform(step_dataframe):
```
patrickherring-TRI, line 1803: Just to be clear, this will only work on maccor? Checking for monotonicity in voltage or capacity does not work? I guess I worry about creating methods this far into pipeline that only work with specific cycler makes since it means we have more to keep track of if/when we have waveforms for Biologic/Arbin

beep/tests/test_structure.py
```diff
@@ -251,6 +254,20 @@ def test_quantity_sum_maccor(self):
             np.all(capacity_sign >= -cycle_sign)
         )  # Capacity increases throughout cycle
 
+    def test_waveform_charge_discharge_capacity(self):
```
patrickherring-TRI, line 257: Nice!

beep/structure.py
```diff
@@ -1073,16 +1081,31 @@ def get_maccor_quantity_sum(data, quantity, state_type):
...
         state_code = MACCOR_CONFIG["{}_state_code".format(state_type)]
-        quantity_agg = data["_" + quantity].where(
-            data["_state"] == state_code, other=0.0, axis=0
-        )
+        quantity_agg = data['_' + quantity].where(data["_state"] == state_code, other=0, axis=0)
+
+        # If a waveform step is present, maccor initializes waveform-specific quantities
+        # that are to be used in place of '_capacity' and '_energy'
+
+        if data['_wf_chg_cap'].notna().sum():
```
patrickherring-TRI, line 1089: I like this. Since it's a method that is specific to maccor already, and we get to use the "built-in" functionality of the system

patrickherring-TRI (changes requested):
Thanks for all of the hard work on this. Know that it was kind of a pain. Just a few changes to make it easier for us to debug and add functionality in the future

beep/structure.py
chirranjeevigopal-TRI: Yeah, I toggled between putting it where it is now vs one level above, i.e in the `get_interpolated_cycles` method, so that the determination is made prior to *any* regular cycle being interpolated and leaving it at a step-level. We could start out with the second option and see. I am writing up a follow-up ticket to see if there is any unexpected behavior with this later on.

beep/structure.py
```diff
@@ -221,7 +221,8 @@ def get_interpolated_steps(
         cycle_indices = self.data.cycle_index.unique()
         cycle_indices = [c for c in cycle_indices if c in reg_cycles]
         cycle_indices.sort()
-        for cycle_index in tqdm(cycle_indices):
+
+        for cycle_index in tqdm(reg_cycles):
```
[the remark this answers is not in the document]
chirranjeevigopal-TRI: This was a version control problem; I had actually set up a filter for `None` scenario, but it seems to have been lost. Will fix.

beep/structure.py
chirranjeevigopal-TRI: For now, it only works for maccor, but can easily add switches for other cyclers if we have sample data.

beep/structure.py
```diff
@@ -304,19 +306,26 @@ def get_interpolated_cycles(
             reg_cycles = [i for i in self.data.cycle_index.unique()]
 
         v_range = v_range or [2.8, 3.5]
+
+        # If any regular cycle contains a waveform step, interpolate on test_time.
+        if self.data[self.data.cycle_index.isin(reg_cycles)].\
+                groupby(["cycle_index", "step_index"]).\
+                apply(determine_whether_step_is_waveform).any():
+            discharge_axis = 'test_time'
```
chirranjeevigopal-TRI, line 314: @patrickherring-TRI this moves the logic for waveform detection one level above, right after splitting out regular and diagnostic cycles.

beep/tests/test_structure.py
```diff
@@ -343,6 +361,29 @@ def test_get_interpolated_charge_step(self):
         self.assertTrue(np.all(np.array(lengths) == 1000))
         self.assertTrue(interpolated_charge["current"].mean() > 0)
 
+    def test_whether_maccor_step_is_waveform(self):
```
patrickherring-TRI, line 364: Could we make this test case a little more complete? Basically, if there were arbin files that failed since voltage was not monotonic it would be good to have them here. I would really like to move away from using the `_wf` columns

beep/structure.py
chirranjeevigopal-TRI: We would need a threshold or some heuristics to determine when a cycle with some non-monotonicities is determined to be waveform vs not.

beep/tests/test_structure.py
patrickherring-TRI: Even just including the Arbin files that didn't pass the monotonicity test in a comment or something would be helpful

beep/conversion_schemas/structured_dtypes.yaml
```diff
@@ -1,4 +1,5 @@
 cycles_interpolated:
+  test_time: 'float32'
```
patrickherring-TRI, line 2: @chirranjeevigopal-TRI Just doing some back of the envelope math, we probably need roughly 10ms resolution even after a year of cycling which means being able to represent ~9 digits in the mantissa? `float32` is a 23bit mantissa which is ~7 digits in the mantissa. So I am guessing this needs to be a float64 if we want to preserve full precision? Kind of wasteful, but I can't think of alternative off the bat. Thoughts?

chirranjeevigopal-TRI: :-) I was calculating with a 100ms resolution, figuring that would already suffice even for regen-braking or HPPC, but just changed it to 64bits.

patrickherring-TRI (approved):
LGTM! Thanks!

PR #132, opened 2020-10-21
Fix arbin metadata requirement

### LINKED ISSUE #122: [Bug] Arbin _Metadata.csv file required even though no column from metadata is critical
## Problem
Instantiating a RawCyclerRun from arbin file without the associated _Metadata.csv file (in this specific naming arrangement) results in a FileNotFound error. There is - seemingly - no reason why this file should be required, as the metadata in the file will never fail anything in the pipeline if it is missing

## Solution
Remove need for _Metadata.csv file, can be incorporated as part of #115

ardunn:
Fix #122 

Previously:
- Arbin metadata file was only nominally required for using RawCyclerRun

Now:
- Arbin metadata is not required, empty metadata is provided if no _Metadata is found
- a warning is thrown via logging to warn user the metadata file is not present

### DIFF
--- a/beep/structure.py
+++ b/beep/structure.py
@@ -723,7 +723,6 @@ class RawCyclerRun(MSONable):
         Returns:
             beep.structure.RawCyclerRun
         """
-        metadata_path = path.replace(".csv", "_Metadata.csv")
         data = pd.read_csv(path)
         data.rename(str.lower, axis="columns", inplace=True)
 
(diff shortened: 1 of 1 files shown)


ardunn:
@patrickherring-TRI only failing check is the coveralls decrease in coverage, which seems inconsequential?

patrickherring-TRI (approved):
LGTM!

PR #133, opened 2020-10-21
Use one authoritative source for requirements

### LINKED ISSUE #113: [Bug] Dependencies have no authoritative source
It can be helpful to have one authoritative source for dependencies.

## Problem

For example, `requirements.txt` does not list `ruamel.yaml==0.16.5` while `setup.py`'s `install_requires` does. This is a minor case, but, if other requirements differ significantly, users might run into discrepancies installing between `pip install beep` (user who wants to use it as written) and `pip install -e . -r requirements.txt` (a developer, or someone who edits the code). This also enables the situation where CI passes on all platforms but users installing the code fail tests due to different dependency versions.

Since other requirements (e.g., `extras_require`) are not listed anywhere, users wanting those extra features (e.g., running tests comprehensively).

## Solution

Use `requirements-*.txt` files as the authoritative source for dependencies. `setup.py` can read these files easily with `pip`'s built in tools:

```
from pip._internal.req import parse_requirements
from pip._internal.network.session import PipSession

pip_requirements = parse_requirements(os.path.join(this_dir, "requirements.txt", PipSession())
reqs = [pii.requirement for pii in p]

setup(name="beep",
      url="https://github.com/TRI-AMDD/beep",
...
      install_requires=reqs,
...

```

Dependencies for CI tests or running the code in production can be similarly enumerated in `requirements-ci.txt` or `requirements-aws.txt` etc. These can be loaded in modular fashion into `setup.py` script.

ardunn:
Requirements text files are now the sole authoritative source for all requirements. They are read automatically using pip and fed into `setup` in `setup.py`.

- The same is done for requirements for testing and documentation.
- Specific versions for testing are specified.
- Docs have been updated to emphasize to users they need the `beep[tests]` extras installed if they want tests to run. 
- Long description markdown is automatically parsed from the readme during setup, including the source and documentation links.
- Closes #113


ardunn:
note to self: need to check the reverts did not undo #132

patrickherring-TRI (approved):
LGTM!

PR #136, opened 2020-10-21
Setup talos

patrickherring-TRI:
This PR is has a test that sets up the Talos project and demonstrates the current state of the scripts.
- csv with Talos parameters
- test that generates the procedures from the csv and then converts these into Arbin schedule files.

Note that the file creation is all done in a scratch directory so the files are not persistent (they can be copied out if they are needed outside of the script)


chirranjeevigopal-TRI (approved):
LGTM!

BruisVanVlijmen2020 (approved):
Looks good to me too

PR #139, opened 2020-10-28
Fix structuring and throughput feature

patrickherring-TRI:
This PR fixes a bug so that the summary attribute no longer includes the diagnostic cycles. It also changes the throughput calculation to separate throughput due to regular and diagnostic cycles.
- Adds the argument to pass the diagnostic_available to the `get_summary` function
- Adds assert statements to a test to ensure that summary no longer contains the diagnostic cycles
- Changes the throughput calculation
- Removes initialization cycle from the throughput calculation
- Adds diagnostic interval information

### DIFF
--- a/beep/dataset.py
+++ b/beep/dataset.py
@@ -33,7 +33,7 @@ from monty.json import MSONable
 from monty.serialization import loadfn, dumpfn
 from functools import reduce
 from beep import MODULE_DIR
-from beep.structure import get_protocol_parameters
+from beep.utils import parameters_lookup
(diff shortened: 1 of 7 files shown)


BruisVanVlijmen2020 (commented):
Looks fine to me

BruisVanVlijmen2020 (approved):

chirranjeevigopal-TRI (approved):
Looks good to me!

pasinger:
Looks good to me too!

PR #140, opened 2020-10-28
MAT-2014 Handles partial success poorly

willpowelson:
Updates `generate_protocol` to handle partial success better.

Now if, say, 28 of 30 possible files are generated we will report:
- The number completed out of the total possible
- The number failed out of the total possible
- A list of the protocol files successfully generated
- A list of the failures to generate a protocol file

### DIFF
--- a/beep/generate_protocol.py
+++ b/beep/generate_protocol.py
@@ -217,6 +217,7 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
         result = "success"
     else:
         message["error"] = "Failed to generate {} of {} protocols".format(num_generation_failures, num_files)
+        logger.error(message["error"])
 
     return successfully_generated_files, file_generation_failures, result, message
 
--- a/beep/generate_protocol.py
+++ b/beep/generate_protocol.py
@@ -82,7 +82,8 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
     # Read csv file
     protocol_params_df = pd.read_csv(csv_filename)
 
-    new_files = []
+    successfully_generated_files = []
+    file_generation_failures = []
     names = []
     result = ""
     message = {"comment": "", "error": ""}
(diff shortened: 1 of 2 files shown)


beep/generate_protocol.py
```diff
@@ -167,27 +168,29 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
...
         logger.info(filename, extra=s)
         if not os.path.isfile(filename):
             protocol.to_file(filename)
-            new_files.append(filename)
+            successfully_generated_files.append(filename)
             names.append(filename_prefix + "_")
 
         elif ".sdu" in template:
+            failure = {
+                "comment": "Unable to find template: " + template,
```
moorepatrick, line 188: Is there a reason not to keep the previous `Schedule file generation is not yet implemented` message?

moorepatrick (commented):
One message question, otherwise LGTM

beep/generate_protocol.py
willpowelson: no, that was a mistake good catch!

beep/generate_protocol.py
```diff
@@ -202,14 +205,20 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
...
+        "error": ""
+    }
     if not result:
         result = "success"
-        message = {
-            "comment": "Generated {} protocols".format(str(len(new_files))),
-            "error": "",
-        }
+    else:
+        message["error"] = "Failed to generate {} of {} protocols".format(num_generation_failures, num_files)
```
patrickherring-TRI, line 219: Would it also make sense send this message to the logger?

patrickherring-TRI (approved):
LGTM

beep/generate_protocol.py
willpowelson: @patrickherring-TRI 
I've never consumed these logs myself so I'm not sure. Right now this appears to be logged in JSON somewhere, I thought that would be sufficient. If you'd like I could add the following line below the one here
```python
logger.error(message["error"])
```
before merging.

patrickherring-TRI: Yeah, if you could add that I think it would provide some better information at the logging level. We are still trying to sort out exactly how to handle logs and other error reporting streams.

PR #141, opened 2020-10-29
Add assert to ensure columns for regular interpolated cycles

patrickherring-TRI:
This PR adds an assert to check that `test_time` is in the regular interpolated cycles


moorepatrick (approved):
LGTM

PR #145, opened 2020-11-04
MAT-2013 error if parent directory does not exist

willpowelson:
Creates parent directory if it does not exist rather than throwing error.


moorepatrick (approved):
LGTM

PR #147, opened 2020-11-06
add seq_num to dataset.data attribute, and expose parameters_path in datasets and features

chirranjeevigopal-TRI:
- PR to add `seq_num` as one of the fields in `BeepDataset.data`.
- When multiple feature-sets are merged, JOIN is performed on both `file` and `seq_num`. This allows runs with potentially same `seq_num` from different projects to be assembled in the same dataset.
- Also handles MAT-2141: exposes `parameters_path` in methods occurring in featurizer_helpers, featurize and dataset modules


patrickherring-TRI (approved):
LGTM

PR #152, opened 2020-11-19
add index for list of dicts key comparison

chirranjeevigopal-TRI:
Adding index for comparing a specific element (a dict) of a list


patrickherring-TRI (approved):
LGTM

PR #154, opened 2020-11-20
serialization and custom hyperparam tests for datasets

chirranjeevigopal-TRI:
PR to add 
- tests for serialization/deserialization of BeepDataset class
- tests for dataset generation from processed_cycler_run objects with custom hyperparam sets. For instance, RPTdQdVFeatures can be generated for the different rpt cycles in the same dataset.

### DIFF
--- a/beep/dataset.py
+++ b/beep/dataset.py
@@ -75,7 +75,7 @@ class BeepDataset(MSONable):
 
     """
 
-    def __init__(self, name, data, metadata, filenames, feature_sets, dataset_dir, missing=None):
+    def __init__(self, name, data, metadata, filenames, feature_sets, dataset_dir='data-share/datasets', missing=None):
 
         """
         Invokes BeepDataset object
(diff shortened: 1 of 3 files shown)


BruisVanVlijmen2020 (changes requested):
With the RPT features that you're generating here, you do realize that the 'charge_y_n' parameter is 1 for all cases? This means that you only take the charge feature for rpt_0.2C, rpt_1C and rpt_2C. 
With the 3 different RPT's the charge rates are in fact all the same: C/5. So the charge features from any 3 RPTs will nearly be the same. The differences are in the discharge portion... So I would suggest turning the charge_y_n parameter to 0.

patrickherring-TRI (approved):
LGTM

BruisVanVlijmen2020 (approved):
Actually, I think it would be great if you could add this line as well:
{'diag_ref': 0, 'diag_nr': 1, 'charge_y_n': 1, 'rpt_type': 'rpt_0.2C', 'plotting_y_n': 0}

So 3 discharges for the three rates, and 1 charge - which are the same across the rates.

LGTM for the rest. 

If you change the other line, I'll approve right away this time :)

chirranjeevigopal-TRI:
@BruisVanVlijmen2020 : this is just the unit-test to ensure dataset generation from processed cycler run works as expected, even when we generate objects of the same featurizer with different  hyperparam combinations. I will, however, use your suggestion for the dataset i generate on the beep-stage instance for building/training EP models

PR #155, opened 2020-11-26
Mat 2094 charging waveforms

patrickherring-TRI:
This PR adds ability to generate waveform charging protocols
- Method to generate stepwise and smooth charging waveforms
- Converter to create the mwf file from the current vs. time waveform
- Generate the procedure and insert the waveform into the correct step
rapid_charge_soc
rapid_charge_matching

Note: Updating the method to include clipping for currents that are too high or too low for the smooth profile. Also, changing the parameterization scheme to three charging rates and two soc points.


beep/protocol/maccor.py
```diff
@@ -1269,6 +1362,79 @@ def insert_driving_parametersv1(reg_params, waveform_directory=None):
...
+        "charge_end_mode_value": 4.25,
+        "discharge_end_mode": "V",
+        "discharge_end_operation": "<=",
+        "discharge_end_mode_value": 2.5,
+        "report_mode": "T",
+        "report_value": 3.0000,
+        "range": "A",
+    }
+    above_80p_c_rate = reg_params["charge_constant_current_2"]
+    soc_initial = 0.05
```
patrickherring-TRI, line 1402: @chirranjeevigopal-TRI initial soc

beep/protocol/maccor.py
```diff
@@ -701,6 +701,99 @@ def generate_procedure_drivingv1(cls, protocol_index, reg_param, waveform_name,
...
+            protocol_index (int): number of the protocol file being
+                generated from this file.
+            reg_param (pandas.DataFrame): containing the following quantities
+
+                charge_percent_limit_1 (float): % of nominal capacity
+                charge_constant_current_2 (float): C
+                charge_cutoff_voltage (float): V
+                charge_constant_voltage_time (integer): mins
+                charge_rest_time (integer): mins
+                discharge_profile (str): {'US06', 'LA4', '9Lap'}
```
patrickherring-TRI, line 719: @patrickherring-TRI Fix the args

beep/protocol/maccor.py
```diff
@@ -1269,6 +1362,79 @@ def insert_driving_parametersv1(reg_params, waveform_directory=None):
     return file_path
 
 
+def insert_charging_parametersv1(reg_params, waveform_directory=None, voltage_limit_offset=0.01, max_c_rate=3.0):
+    """
+    Args:
+        reg_params (pandas.DataFrame): containing the following quantities
+            charge_type_1 (str): {'smooth', 'step'}
+            charge_current_param_1 (float): charge c-rate for first 20%
```
patrickherring-TRI, line 1370: @patrickherring-TRI Update comments (equally distributed windows over the charging range)

beep/utils/waveform.py
```diff
@@ -56,3 +58,223 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+        mesh_points_mid[0:-1] += 1 / (len(mesh_points_mid)) * 0.5
+
+        mesh_points_mid = list([self.soc_i]) + list(mesh_points_mid) + list([self.soc_f * 1.01])
+        mesh_points_mid = np.array(mesh_points_mid)
+
+        charging_c_rate_start = self.final_c_rate
+        rates = np.array([charging_c_rate_start] + charging_c_rates + [self.final_c_rate] + [self.final_c_rate])
+        interpolator = interpolate.PchipInterpolator(mesh_points_mid, rates, axis=0, extrapolate=0)
+
+        charging_c_rate_soc1_end = interpolator.__call__(mesh_points[1])
```
patrickherring-TRI, line 233: @pasinger Should this be `mesh_points[-1]`? If so, then some of the smooth curves change substantially when we do the time matching to 80%

pasinger: I don't think it should be `mesh_points[-1]`. This value is a part of the routine we had decided on for how to set the C rates in the first SOC window. There was some ambiguity originally because we have the C rates "clamping" at the mid-points of the SOC windows, and it would be unclear what behavior should be set for the first C rate before it's mid-point. Obtaining the C rate value from `mesh_points[1]` is part of the routine to either a) set the slope to constant between the first two SOC points or b) set the C rate before the first SOC mid-point to be CC at its specified value. These are calculated in `charging_c_rate_start` below. Case (a) occurs if the slope will be negative and Case (b) occurs if the slope would have been positive.
It may be worth revisiting this behavior since we have added constraints into the new parameterization.

pasinger (changes requested):
Overall looks really good to me. Thanks for implementing all of this!
There are a few pretty minor changes that should be made before approval, and at least one aspect we should review / give further thought to as well.

beep/utils/waveform.py
```diff
@@ -150,11 +155,11 @@ def get_input_currents_both_to_final_soc(self, charging_c_rates):
                                maxfev=1000,
                                factor=0.1,
                                epsfcn=0.01,
-                               args=(end_soc_time, charging_c_rates))
+                               args=(end_soc_time, charging_c_rates, mesh_points))
 
         charging_c_rates = list(np.add(charging_c_rates, offset_solved))
```
pasinger, line 160: Change line to
charging_c_rates = [charging_c_rates[0]] + list(np.add(charging_c_rates[1:-1],offset)) + [charging_c_rates[-1]]
This change makes it so that only the non-edge c_rate values are offset.

pasinger (changes requested):
Make additional modifications to algorithm around the offsetting to ensure desirable behavior.

beep/utils/waveform.py
```diff
@@ -56,3 +58,244 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+            offset (float): amount to shift the current values
+            charging_c_rates (list): c-rates for each of the charging steps. Each step is assumed to be an equal
+                SOC portion of the charge, and the length of the list just needs to be at least 1
+            mesh_points (list): soc values for beginning and end of each of the charging windows
+
+        returns
+        np.array: array with the current as a function of soc
+        np.array: array with the corresponding soc values
+        """
+        shifted_c_rates = list(np.add(charging_c_rates, offset))
```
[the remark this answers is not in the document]
patrickherring-TRI: @pasinger I am having some trouble with this change. A bunch of tests are failing since it can't seem to get to the same charge time...

beep/protocol/maccor.py
```diff
@@ -715,7 +715,8 @@ def generate_procedure_chargingv1(cls, protocol_index, reg_param, waveform_name,
                 charge_current_param_1 (float): c-rate for first charging window
                 charge_current_param_2 (float): c-rate for second charging window
                 charge_current_param_3 (float): c-rate for third charging window
-                charge_current_param_4 (float): c-rate for fourth charging window
+                charge_soc_param_1 (float): soc upper bound for first charging window
+                charge_soc_param_2 (float): soc upper bound for second charging window
                 charge_percent_limit_1 (float): % of nominal capacity to end fast charging
```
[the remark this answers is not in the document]
patrickherring-TRI: Done

beep/protocol/maccor.py
```diff
@@ -1366,7 +1367,7 @@ def insert_driving_parametersv1(reg_params, waveform_directory=None):
     return file_path
 
 
-def insert_charging_parametersv1(reg_params, waveform_directory=None, max_c_rate=3.0):
+def insert_charging_parametersv1(reg_params, waveform_directory=None, max_c_rate=3.0, min_c_rate=0.2):
     """
     This function generates the waveform file for rapid charging. The charging windows are
     equally divide up the soc window (defined by charge_start_soc_1 to charge_percent_limit_1)
```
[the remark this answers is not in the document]
patrickherring-TRI: Changed this name in the spreadsheet and in the code.

beep/protocol/maccor.py
```diff
@@ -1406,10 +1409,14 @@ def insert_charging_parametersv1(reg_params, waveform_directory=None, max_c_rate
     soc_initial = reg_params["charge_start_percent_1"] / 100
     soc_final = reg_params["charge_percent_limit_1"] / 100
     charging_c_rates = [reg_params["charge_current_param_1"], reg_params["charge_current_param_2"],
-                        reg_params["charge_current_param_3"], reg_params["charge_current_param_4"]]
-    above_80p_c_rate = max([charging_c_rates[-1] * 0.8, reg_params["charge_constant_current_2"]])
-    charging = RapidChargeWave(above_80p_c_rate, soc_initial, soc_final, max_c_rate)
-    current_smooth, current_step, time_uniform = charging.get_currents_with_uniform_time_basis(charging_c_rates)
+                        reg_params["charge_current_param_3"], reg_params["charge_constant_current_2"]]
```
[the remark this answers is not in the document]
patrickherring-TRI: Changed this back to "charge_current_param_4" to match the other values

beep/utils/waveform.py
```diff
@@ -56,3 +58,244 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+
+        return current_smooth_time_adjusted, time_smooth_time_adjusted, current_multistep, time_multistep
+
+    def offset_value(self, offset_test, *data):
+        """
+        Helper function shift the charging rates for the smooth current to ensure that both the multistep and
+        smooth charging functions reach the end in the same length of time.
+
+        Args:
+            offset_test (float): Amount to shift all of the c rates in the smooth curve
```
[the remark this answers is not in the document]
patrickherring-TRI: good catch!

beep/utils/waveform.py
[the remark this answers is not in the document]
patrickherring-TRI: Good catch, thanks!

beep/utils/waveform.py
patrickherring-TRI: changed this one too

beep/utils/waveform.py
patrickherring-TRI: Yeah, I think we can leave the behavior as is for now. I can't really think of a good alternative off-hand.

beep/utils/waveform.py
```diff
@@ -56,3 +58,246 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+            max_c_rate (float): maximum charging rate for the charge
+            min_c_rate (float): minimum charging rate for the charge
+
+        """
+        self.final_c_rate = final_c_rate
+        self.soc_i = soc_initial
+        self.soc_f = soc_final
+        self.max_c_rate = max_c_rate
+        self.min_c_rate = min_c_rate
+        self.soc_points = 1000
```
chirranjeevigopal-TRI, line 90: Can we make `soc_points` a global/config variable (if it is not going to be passed as an argument to the constructor)?

beep/utils/waveform.py
```diff
@@ -56,3 +58,246 @@ def convert_velocity_to_power_waveform(waveform_file, velocity_units):
...
+            mesh_points (list): soc values for beginning and end of each of the charging windows
+
+        returns
+        np.array: array with the adjusted smooth current as a function of uniform time basis
+        np.array: array with the multistep current as a function of the uniform time basis
+        np.array: array with the corresponding uniformly spaced time values
+        """
+        current_smooth, time_smooth, current_multistep, time_multistep = \
+            self.get_input_currents_both_to_final_soc(charging_c_rates, mesh_points)
+        end_time_smooth = np.int(np.round(np.max(time_smooth), 0))
```
chirranjeevigopal-TRI, line 108: Would it help to assert that `mesh_points` (soc_list) is monotonically increasing?

beep/protocol/maccor.py
```diff
@@ -1269,6 +1369,87 @@ def insert_driving_parametersv1(reg_params, waveform_directory=None):
     return file_path
 
 
+def insert_charging_parametersv1(reg_params, waveform_directory=None, max_c_rate=3.0, min_c_rate=0.2):
```
chirranjeevigopal-TRI, line 1372: Minor comment: what's the reason for `v1` designation for this method?

beep/utils/waveform.py
patrickherring-TRI: I don't think that making it a global variable is a good idea here since we might need to alter the interpolation. Is there a specific reason you were thinking it should be a global variable?

beep/protocol/maccor.py
patrickherring-TRI: This is a specific way of parameterizing the fast waveform. Since there might be others down the line, I am hedging a bit to allow for new functions that allow for different parameterizations. Usually for these functions, we want legacy functions to continue to work in the same way so that we can reproduce older results.

beep/utils/waveform.py
patrickherring-TRI: Good idea, adding that in

beep/utils/waveform.py
chirranjeevigopal-TRI: No strong reason -- I wasn't sure if having a hardcoded variable buried in was better than having it defined as a config/global variable when we needed to change. But, since its a class variable, I suppose its ok!

chirranjeevigopal-TRI (approved):
Thanks for making these significant additions. LGTM!

pasinger (approved):
looks good to me!

BruisVanVlijmen2020 (approved):
Trusting your wisdom

pasinger (approved):
looks good!

PR #157, opened 2020-12-07
Mat 1973

willpowelson:
A first pass at converting Maccor Protocol files to Biologic Procedure files with a single Modulo Bat Technique.

Allows users to convert directly between files, or load up an AST of the Maccor XML, alter it, and convert the altered AST into a file.

This is particularly important because the file we wish to convert, `diagnosticV4.000` has multiple steps containing EndEntries (conditions on which a step ends, and a GOTO is executed) that result in a protocol not representable in Biologic Format. Seee `test_remove_end_entries_by_pred` to see an example of what this might look like using the alteration condition we need to make `diagnosticV.4000` convertible. 

### Conversion Details:
Limitations of the Biologic File Format:

1.  There is no "Advance Cycle" step type
2.  Cycle number only advances based on a file defined `Cycle Definition` has two values
    - `Loop` which advances when a loop triggers
    - `Charge/Discharge Alternance`
3. Loops cannot be nested
4. Metric values can only have 3 decimals of precision, and a max value of 10,000,000
5. GOTO statements cannot GOTO before a a loop, or before a loop, and have that loop run more than once
6. The maximum number of steps in biologic file, while not known precisely, is at most 9,999

From these limitations 3 techniques arrive for conversion
1. We create what is effectively an "Advance Cycle" step with 2 loop steps. With cycle definition set to `Loop`, if we add a loop step that loops 0 times (a "blank" loop), followed by a loop step that loops exactly once to the "blank" loop will cause the cycle to advance. Note that due to limitation (3) these advance cycles cannot occur in a loop, more on that in a moment.
2. We can "unroll" loops to make them representable. The conditions for a loop to unroll  follow from limitation 3. and the fact that we are using the `Loop`  definition of a cycle.
    - Any loop containing a loop must unroll (3) or else the experiment will be altered
    - Any loop containing an advance cycle that does not immediately precede the loop step must unroll or else the cycle numbers will be incorrect
    - Any loop that does not have an advance cycle not immediately preceding the loop step must unroll or else the cycle numbers will be incorrect
3. We are conservative about unrolling both to keep files readable, and in an attempt to not pass the max number of steps allowed in a Biologic file
4. We make a "best effort" conversion of metrics so that they are both readable and representable without losing any significant figures.

I've left in print statements about the number of unrolled steps generated in the conversion process if you'd like to test the conversion. I'd recommend the following code with relevant filepaths set appropriately for `diagnosticV4.000`
```
def pred(end_entry, step_num):
    goto_step = int(end_entry["Step"])
    # filter all goto step 70s, except when that is Next Step
    return  goto_step != 70 or step_num == 69

converter = MaccorToBiologicMb()
maccor_ast = converter.load_maccor_ast(fp)
filtered_ast = converter.remove_end_entries_by_pred(maccor_ast, pred)
conveter.maccor_ast_to_protocol_file(filtered_ast, out_path)
```
I'd appreciate any attempts that could be made for a converted file to be loaded onto other versions of EC-Lab software to detect any bugs I may have missed.


beep/tests/test_maccor_to_biologic_mb.py
```diff
@@ -55,14 +142,14 @@ def test_rest_step_conversion(self):
...
-            "          <Step>070</Step>"
+            "          <Step>02</Step>"
             "          <Value>4.4</Value>"
             "        </EndEntry>"
             "        <EndEntry>"
             "          <EndType>Voltage </EndType>"
             "          <SpecialType> </SpecialType>"
             "          <Oper>&lt;= </Oper>"
-            "          <Step>070</Step>"
+            "          <Step>02</Step>"
```
patrickherring-TRI, line 152: I think this should be `002` The step number is zero padded to 3 digits

beep/tests/test_maccor_to_biologic_mb.py
```diff
@@ -159,34 +246,428 @@ def test_discharge_current_step_conversion(self):
...
+            "Characteristic mass : 9.130 mg\r\n"
+            "Cycle Definition : Loop\r\n"
+            "Turn to OCV between techniques\r\n"
+            "\r\n"
+            "Technique : 1\r\n"
+            "Modulo Bat\r\n"
+            "Ns                  0                   1                   2                   3                   4                   5                   6                   7                   8                   \r\n"
+            "type                Rest                CC                  CV                  CV                  CV                  CC                  Loop                Loop                Loop                \r\n"
+            "Apply I/C           I                   C / N               C / N               C / N               C / N               C / N               C / N               C / N               C / N               \r\n"
+            "ctrl1_val                               1.000               3.300               3.300               3.300               142.900             100.000             100.000             100.000             \r\n"
```
patrickherring-TRI, line 563: Some of these control values don't look quite right given the procedure values. Type `CC` should correspond to StepMode `Current` and Type `CV` should correspond to StepMode `Voltage` I think the Limit values from the maccor are a little tricky, but they should probably go into `ctrl2_val` and `ctrl2_val_unit`
Let me know if that makes sense?

patrickherring-TRI (changes requested):
Had some fixes to the way that the steps with limits are being translated. Let me know if it would help to explain in more detail!

beep/tests/test_maccor_to_biologic_mb.py
willpowelson: good catch

beep/tests/test_maccor_to_biologic_mb.py
willpowelson: I will make the changes for `Limits` then.

`Current` corresponds to `CC` and `Voltage` to CV you say. This is the output I'm expecting, if that's unclear, let me break down the conversion:

1. Step 1 is `Rest` so seq 0 is `Rest`
2. Step 2 is `Current` so seq 1 is `CC`
3. Step 3 is `Do 1` which is not representable as a seq
4. Step 4 is `Voltage` so seq 2 is `CV`
5. Step 5 is `Loop 1` paired with step 3, it's value is 2 which means all steps between it and the opening do will occur 2 more times for a total of 3 occurrences. Since there is no advance cycle at the end of the loop, the loop must unroll. This means seq 3 and seq 4 are `CV`, copies of seq 2.
6. Step 6 is `Do 1` which is not representable as a seq
7. Step 7 is `Current` so seq 5 is `CC`
8. Step 8 is  `Adv Cycle`,  it fulfills the conditions for the containing loop not to unroll and so is not converted to a seq.
9. Step 9 is `Loop 1`, it does not unroll so seq 6 is `Loop` and loops to seq 5.
10. Step 10 is `Adv Cycle`, it is not in a loop, so the functionality is mocked by adding 2 loop steps as documented in the PR desciption above, these are seqs 7 and 8.

beep/tests/test_maccor_to_biologic_mb.py
```diff
@@ -159,34 +246,428 @@ def test_discharge_current_step_conversion(self):
...
+            "      <Reports/>"
+            "      <Range></Range>"
+            "      <Option1></Option1>"
+            "      <Option2></Option2>"
+            "      <Option3></Option3>"
+            "      <StepNote></StepNote>"
+            "    </TestStep>"
+            "    <TestStep>"
+            "      <StepType> Charge </StepType>"
+            "      <StepMode>Voltage </StepMode>"
```
patrickherring-TRI, line 370: I think that if the limit is voltage based, StepMode needs to be `Current` or vise-versa

moorepatrick:
@willpowelson What's the status of this?

patrickherring-TRI (approved):
LGTM, lets merge and work on specific issues. Thanks for all of the hard work!

willpowelson:
@moorepatrick
Sorry for not responding, there seems to have been a problem with the way my e-mail notification settings and I never saw the mention. This got put on hold but we're now merging.

PR #158, opened 2020-12-10
Change the exit step for safety to avoid an infinite loop 

patrickherring-TRI:
This PR adds a template for the PreDiag project. The template avoids a problem with the earlier version where exiting a step due to a safety condition in the final diagnostic could lead to an infinite loop.


chirranjeevigopal-TRI (approved):
lgtm!

PR #159, opened 2020-12-17
Mat 2150 threshold calculation

patrickherring-TRI:
This PR adds a function to calculate the cycle index and throughput at which the run drops below a threshold for the remaining fractional capacity.
- Function to calculate the values for the threshold targets for early prediction
- Ability to extrapolate for cells that have not yet reached the threshold
Screen Shot 2020-12-17 at 12 35 06 AM


beep/dataset.py
```diff
@@ -339,6 +340,104 @@ def serialize(self):
         return self.dataset_dir
 
 
+def get_threshold_targets(dataset_diagnostic_properties,
+                          cycle_type_target="rpt_1C",
```
chirranjeevigopal-TRI, line 344: Just to be consistent with the naming (with other modules), can we name this `cycle_type`?

beep/dataset.py
```diff
@@ -339,6 +340,104 @@ def serialize(self):
...
+                          basis="discharge_energy",
```
chirranjeevigopal-TRI, line 345: Again, for the sake of consistency with other modules and method, call `basis` ->`metric`?

beep/dataset.py
```diff
@@ -339,6 +339,87 @@ def serialize(self):
...
+                          threshold=0.80,
+                          filter_kinks=None,
+                          extrapolate_threshold=True):
```
BruisVanVlijmen2020, line 348: If not `None`, what could `filter_kinks` be for example?

patrickherring-TRI: I found for some of the cells a value of 0.04 seemed to be a good choice. Obviously depends a lot on what conditions you are trying to filter out.

beep/dataset.py
patrickherring-TRI: yup, done!

beep/dataset.py
patrickherring-TRI: done!

chirranjeevigopal-TRI (approved):
lgtm!

BruisVanVlijmen2020:
looks good to me too

PR #160, opened 2020-12-22
Hotfix remove current value assertion

patrickherring-TRI:
This PR removes an assertion that has been deemed unnecessary.


moorepatrick (approved):
LGTM

PR #161, opened 2021-01-06
Mat 2107 add step index inference

patrickherring-TRI:
This PR adds a function to dynamically determine the step indices for the hppc diagnostic cycle. This is critical for cases where the order and number of steps is not strictly preserved from between all of the runs in the project.
- Works for cycle type `hppc`
- Determines what each step is based on average current, duration of the step and time based location in the cycle
- Can be expanded to cover the other diagnostic cycles
- Probably needs to be expanded/modified to cover the diagnostic cycles from the Biologic


chirranjeevigopal-TRI (approved):
LGTM!

PR #162, opened 2021-01-07
Change the validation voltage limits

patrickherring-TRI:
This PR changes the voltage limits for validation to allow processing of files where the cell has been disconnected at the end of the test so the voltage floats.
- Change upper limit to 5V
- Change lower limit to 0V

Tested with "xIris_000002_0000B2_CH3.csv" and "PreDiag_000298_000159.008" to make sure structuring succeeds and these files are now marked as valid.


chirranjeevigopal-TRI (approved):
lgtm!

PR #163, opened 2021-01-07
Fix negative paused values

chirranjeevigopal-TRI:
Change the datatype for `paused` summary attribute to be `int32`.  Since `paused` evaluates max pause duration in seconds, `int8` type was leading to negative values anytime 127 was exceeded.


patrickherring-TRI (approved):
LGTM

PR #165, opened 2021-01-21
MAT-2195 Use of regular cycles in DiagnosticSummaryStats feature class

JosephMontoya-TRI:
This PR is a draft of some initial functionality for the use of regular cycles in the DiagnosticSummaryStats feature class.  

* Adds hyperparameters `use_regular_cycles` and `regular_cycle_comp_num` to the class in order to trigger regular cycles for features and select comparison cycles
* Modified the fetching of data from the processed_cycler_run dataframe accordingly.

I think it's still a bit rough around the edges and could use some feedback, specifically:
* Would it be better to create a new class that creates these features separately?  I'm not sure whether this would be useable, but it seems like the switch logic makes a lot of the parameters implicit and might be confusing to users, rather than just having a whole separate class.
* I don't have a test file that includes discharge/charge energy, so the test is currently failing, should I revert the energy features to diagnostic or is there an example we can use that includes energy columns for the regular cycles?
* I'm not 100% sure the refactor I've done which replaces the list comprehension and zip logic (e.g. `QcDiff = [a_i - b_i for a_i, b_i in zip(Qc100_1, Qc10_1)]` with the equivalent numpy operations is okay from a memory standpoint.  I have a vague memory that at some point we refactored some of these things to use more basic python types to save on memory and just want to make sure I'm not reverting anything that shouldn't be.


patrickherring-TRI:
@JosephMontoya-TRI Sorry for the delay, just getting back to this after a crazy week
Regarding your questions:
1. I do think it would be better to use a different class just for clarity reasons. Featurization is very prototype-y so the less dependency between features the better I think
2. We might need to add the charge/discharge energy in the structuring step since I don't think it's computed right now. I can break that out into another ticket or if you want to handle it here, we can do that.
3. Memory should not be a problem, featurization is a much lower memory footprint

JosephMontoya-TRI:
@patrickherring-TRI this is ready for re-review

beep/featurize.py
```diff
@@ -595,17 +599,184 @@ def features_from_processed_cycler_run(cls, processed_cycler_run, params_dict=No
...
+                            # 'discharge_energy'
+                            ]
+        pcycler_run_columns = processed_cycler_run.cycles_interpolated.columns
+        if not all([column in pcycler_run_columns for column in required_columns]):
+            return False
+
+        return True
+
+    STANDARD_OPERATION_NAMES = ["var", "min", "mean", "skew", "kurtosis", "abs", "square"]
+
```
chirranjeevigopal-TRI, line 669: Suggest renaming this to `SUMMARY_STATISTICS` or something like it. Operation is a little over-loaded with physical battery operation :-)

beep/featurize.py
```diff
@@ -595,17 +599,184 @@ def features_from_processed_cycler_run(cls, processed_cycler_run, params_dict=No
...
+    @staticmethod
+    def get_standard_operation_values(array):
+        """
```
chirranjeevigopal-TRI, line 672: Consistent with the change above, maybe rename this method too to say `get_summary_statistics`?

chirranjeevigopal-TRI: I do like that you broke it out into a separate method

chirranjeevigopal-TRI:
But for the minor changes, lgtm. Thank you!

beep/featurize.py
JosephMontoya-TRI: Agreed, used SUMMARY_STATISTIC_NAMES

beep/featurize.py
JosephMontoya-TRI: agreed, renamed to `get_summary_statistics`

beep/featurize.py
```diff
@@ -676,38 +849,13 @@ def features_from_processed_cycler_run(cls, processed_cycler_run, params_dict=No
...
 
         # Determine beginning and end of investigated cycle type
-        index_pos_list = [
-            i
-            for i in range(len(diagnostic_interpolated.cycle_type))
-            if diagnostic_interpolated.cycle_type[i]
-            == params_dict["diagnostic_cycle_type"]
-        ]
+        index_pos_list = np.where(
+            diagnostic_interpolated.cycle_type == params_dict["diagnostic_cycle_type"])[0]
```
patrickherring-TRI, line 853: Would it be better to use `np.asarray(condition).nonzero()`?

beep/featurize.py
```diff
@@ -718,16 +866,10 @@ def features_from_processed_cycler_run(cls, processed_cycler_run, params_dict=No
         Qc10_1 = diagnostic_interpolated.charge_capacity[
                  start_list[params_dict["cycle_comp_num"][0]]:
                  start_list[params_dict["cycle_comp_num"][0]] + params_dict["Q_seg"]]
```
patrickherring-TRI, line 868: Could you add a TODO here? I realized that the `"Q_seg"` value is really just the number of interpolated datapoints for these diagnostic cycles so if we ever change that, this value will also need to be changed.

beep/tests/test_features.py
```diff
@@ -29,6 +29,7 @@
     HPPCRelaxationFeatures,
     DiagnosticProperties,
     DiagnosticSummaryStats,
+    CycleSummaryStats
```
patrickherring-TRI, line 32: Awesome, thanks for remembering this!

patrickherring-TRI (approved):
LGTM

beep/featurize.py
JosephMontoya-TRI: done

beep/featurize.py
JosephMontoya-TRI: I personally like np.where because I think it's more readable, but I see in the numpy docs that they suggest calling nonzero directly, so I changed it.

PR #167, opened 2021-02-01
Initial refactor of the protocol generation logic

patrickherring-TRI:
This PR is an attempt to make protocol generation more intuitive and easier to modify.
- Any template with the right number of steps will be parameterized similarly - new templates with minor modifications are fine
- Additional decision logic can be added for more specific cases
- protocol files can be overwritten now


beep/generate_protocol.py
```diff
@@ -89,96 +108,88 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
...
-            )
-            diagnostic_params = diag_params_df[
-                diag_params_df["diagnostic_parameter_set"]
-                == protocol_params["diagnostic_parameter_set"]
-            ].squeeze()
+        if ".000" in template:
             template_fullpath = os.path.join(PROCEDURE_TEMPLATE_DIR, template)
-            if protocol_params["project_name"] == "RapidC":
+            template_length = template_detection(template_fullpath)
+            if "diagnostic_parameter_set" in protocol_params:
```
moorepatrick, line 125: Can we put in some comments next to each of these `if`'s and `elif`'s to describe what we are matching.

beep/generate_protocol.py
```diff
@@ -89,96 +108,88 @@ def generate_protocol_files_from_csv(csv_filename, output_directory=None):
...
+                file_generation_failures.append(failure)
+                warnings.warn("Unsupported file template {}, skipping.".format(template))
+                result = "error"
+                continue
+
             filename = "{}.000".format(filename_prefix)
             filename = os.path.join(output_directory, "procedures", filename)
-        elif template == "formationV1.mps":
+
+        elif ".mps" in template and template == "formationV1.mps":
```
moorepatrick, line 179: Some more comments next to these would be helpful too.

moorepatrick (changes requested):
LGTM but a few more comments would be helpful.

beep/generate_protocol.py
patrickherring-TRI: Good idea! Added those now

beep/generate_protocol.py
patrickherring-TRI: Yup, adding comments to each of the cases

moorepatrick (approved):
LGTM

chirranjeevigopal-TRI (approved):
LGTM!

PR #172, opened 2021-02-24
Mat 2360 add rpt and reset to get step index

patrickherring-TRI:
This PR adds support for rpt and reset cycles to the get_step_index function


chirranjeevigopal-TRI (approved):
lgtm!

PR #173, opened 2021-02-25
Bugfix - Add filtering to features to remove appended cycles

patrickherring-TRI:
This PR addresses an issue for some of the data where a final diagnostic was appended to the end of the file with a cycle number that matched the first diagnostic. However since the test_time correctly keeps increasing it is possible to filter out those cycles
- Adds filtering to all of the Featurizer classes
- Reconfigures the selection of the diagnostic charge and discharge steps to use newer, built in functions


pasinger (approved):
LGTM

chirranjeevigopal-TRI (approved):
LGTM!

PR #175, opened 2021-02-27
Bugfix change diagnostic cycle determination

patrickherring-TRI:
This PR changes the method for finding diagnostic cycles. Instead of assuming that diagnostic cycles are occurring at specific cycle numbers, it is using the location of the first two diagnostics in the test to get the step numbers of the diagnostics and then looking for those step numbers throughout the cycles. Wherever those step numbers occur, that cycle is marked as the beginning of a diagnostic. 
- Gets step numbers for reset cycle in the diagnostic
- Asserts that the index for those cycles is contiguous
- Marks any subsequent cycles that contain either of those sets of steps as the beginning of a set of diagnostic cycles


beep/structure.py
```diff
@@ -1235,39 +1235,27 @@ def determine_structuring_parameters(
...
                 v_range = [
                     all_parameters["discharge_cutoff_voltage"].min(),
                     all_parameters["charge_cutoff_voltage"].max(),
                 ]
-            if {
-                "diagnostic_type",
-                "diagnostic_start_cycle",
-                "diagnostic_interval",
-            }.issubset(run_parameter.columns):
+            if {"diagnostic_type", "diagnostic_start_cycle", "diagnostic_interval", }.issubset(run_parameter.columns):
```
moorepatrick, line 1243: Do we need the extra `, ` after `diagnostic_interval`?

beep/structure.py
```diff
@@ -1235,39 +1235,27 @@ def determine_structuring_parameters(
...
-                                run_parameter["diagnostic_interval"].iloc[0]
-                                + hppc_rpt_len
-                            )
-                            + 1
-                            + run_parameter["diagnostic_start_cycle"].iloc[0]
-                            + 1 * hppc_rpt_len
-                        )
-                        diagnostic_starts_at.append(diag_cycle_num)
+                    initial_diagnostic_at = [1, 1 + run_parameter["diagnostic_start_cycle"].iloc[0] + 1 * hppc_rpt_len]
+                    assert (np.diff(self.data[self.data.cycle_index == initial_diagnostic_at[0]].index) < 2).all(), \
```
moorepatrick, line 1248: @patrickherring-TRI can you explain what's going on here? 
`np.diff(self.data[self.data.cycle_index == initial_diagnostic_at[0]].index) < 2` 

That looks like `np.diff(self.data[ boolean ].index) < 2` which doesn't make any sense.

moorepatrick (commented):
I'm not familiar with a lot of what's going on in the main chunk of the changes. Can you explain what's going on?

beep/structure.py
patrickherring-TRI: So this is meant to check that the index for the selected cycle is continuous (ie. incrementing by one). The reason to check for this is that a number of PredictionDiagnostics files have the same cycle number at the end of the file where a set of diagnostics was rerun. This should not be happening anymore, and ideally this problem would get caught in the data validator. To solve this issue, it might actually make more sense to reprocess these files and check that they get marked as invalid. Then I can remove this portion of the fix.

chirranjeevigopal-TRI (approved):
LGTM!

beep/structure.py
```diff
@@ -1235,39 +1235,38 @@ def determine_structuring_parameters(
...
-                            + 1
-                            + run_parameter["diagnostic_start_cycle"].iloc[0]
-                            + 1 * hppc_rpt_len
-                        )
-                        diagnostic_starts_at.append(diag_cycle_num)
+                    initial_diagnostic_at = [1, 1 + run_parameter["diagnostic_start_cycle"].iloc[0] + 1 * hppc_rpt_len]
+                    # Calculate the number of steps present for each cycle in the diagnostic as the pattern for
+                    # the diagnostic. If this pattern of steps shows up at the end of the file, that indicates
+                    # the presence of a final diagnostic
+                    diag_0_pattern = [len(self.data[self.data.cycle_index == x].step_index.unique()) for x in
```
patrickherring-TRI, line 1251: @viveklam I think this will be able to find the final diagnostic in the file. It creates a list of the diagnostic cycles (reset, hppc, etc.) with the number of steps for each cycle. Then it looks at the last ~5 cycles in the file and if they have the same pattern of number of steps for those final cycles, it marks the those as belonging to a diagnostic.

patrickherring-TRI: ie. diag_0_pattern = [2, 7, 2 ,2, 2]

moorepatrick (approved):
LGTM

PR #177, opened 2021-03-04
Change pandas indexing usage to use iloc instead of straight lists

patrickherring-TRI:
This PR updates the DeltaQFastCharge feature class to use standard pandas indexing selection instead of lists. Output of the feature class is unchanged in tests.
- Passes index lists to `.iloc` which has the required behavior
- Add condition for creating the feature class to ensure basic indexing columns are present


chirranjeevigopal-TRI (approved):
LGTM!

PR #179, opened 2021-03-13
MAT-2259 Add summary stats to features

patrickherring-TRI:
This PR adds differences in the diagnostic summary stats to the DiagnosticSummaryStats feature class.
- Adds function that calculates the difference between the first and second diagnostic for several different metrics
- Turns on calculation of charge features for the regular cycles


moorepatrick (approved):
LGTM

PR #190, opened 2021-03-15
MAT-2382 Dependabot

JosephMontoya-TRI:
Adds dependabot.yml and initial dependabot updates (from my fork).

Note that I accidentally merged these earlier, but reverted the changes.  Somehow the earlier dependency updates got lost in the reversion, so they're here.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,3 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "pip"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/"
+    # Check the pip registry for updates once a week
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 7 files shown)


JosephMontoya-TRI:
Debugging the tests.  There's an issue with the latest version of pydash that i'm going to punt on because it looks like an API change, so I've pinned it to the last version we used for now (dependabot will issue a PR, which we can get to when we have time).

For the remaining issue, Scipy is the root culprit (it works with scipy 1.4.1 but fails with scipy 1.6.1), but I'm not sure why yet.  This is a little more fundamental of a dependency, so I think we should address now. Will update when I know more.

JosephMontoya-TRI:
It's from a soft change from Scipy 1.5.4->1.6.1 that probably has something to do with interp1d's implementation of `slinear`

patrickherring-TRI:
@JosephMontoya-TRI yes, I think the latest version of scipy that I could get to work was 1.5.1 Has `slinear` been deprecated in favor of `linear`?

JosephMontoya-TRI:
No slinear still exists and is the same thing according to the documentation, but the interpolation seems to have different behavior.  Not yet sure what the differences are (the max is still the same, for example) so it may not be related to `slinear`, but some other housekeeping thing interp1d is doing.  I'll try a comparison of the output between the versions.

JosephMontoya-TRI:
So I think the test is passing, but coveralls throwing an error.

patrickherring-TRI:
@JosephMontoya-TRI Thanks for getting dependabot setup. So any time a new version of a package is issued, dependabot will check it out and issue a PR?

JosephMontoya-TRI:
Yep, I have it set to a weekly cadence in the config file, so it should issue PRs on monday.  If we want these to be less frequent, we can configure.  Also, there's a bit of an issue with botocore/boto3, which sometimes need to be updated together to install successfully, I think eventually these will be able to be coupled by configuring dependabot, but not a supported feature yet.

Re: scipy, I was in the process of comparing the interpolated values in the failing test, I just installed one, invoked the debugger, installed the other, and invoked the debugger again in a separate session so I could poke around (you could also install them in different envs).  I found that the differing values are around 3143 in the two sets, but most of the other values are exactly the same. I suspect there's a smoothing parameter in the linear spline for which the default was increased which will have an impact. We may want to just fall back on `linear` as the interpolation method, which shouldn't be parameterized at all, but this could change a few things, so we'll probably want to test that a little to get a sense of the impact.

JosephMontoya-TRI:
FYI dependabot should issue the PRs for scipy and for pydash now I think

PR #196, opened 2021-03-19
Change environment variable for big tests

patrickherring-TRI:
Fix environment variable so that big tests run

### DIFF
--- a/.github/workflows/beep-test.yml
+++ b/.github/workflows/beep-test.yml
@@ -39,7 +39,7 @@ jobs:
         env:
           MPLBACKEND: "Agg"
           BEEP_ENV: "dev"
-          BEEP_BIG_TESTS: "True"
+          BIG_FILE_TESTS: "True"
         run: |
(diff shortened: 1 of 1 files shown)


danielschweigert-TRI (approved):

ardunn:
@patrickherring-TRI does this mean the bigfile tests are only run on merges?


`BIG_FILE_TESTS` is only set in `beep_test.yml`, not `beep-test-pr.yml`. So they appear to only run on merges. Just want to confirm this is the intended behavior.

patrickherring-TRI:
@ardunn Yes, it has to do with the permissions for PRs. Basically, since anyone can add anything to a PR, PRs are not allowed to access any secrets (in case the PR somehow sends that secret someplace). Thus the file can't be retrieved from S3 and the test can't run. We might fix this in the future by just making the file public so that no access key is required...

ardunn (approved):

PR #221, opened 2021-03-30
Add service information into the base class for the logger

patrickherring-TRI:
Add necessary information for the logger in the base class. This addresses an error in the data pipeline.


ardunn:
Thanks for doing this @chirranjeevigopal-TRI , I must have accidentally removed the service config when writing BEEPDatapath

chirranjeevigopal-TRI:
I will pass the good word to @patrickherring-TRI :-)

PR #235, opened 2021-04-13
refactor DashOrderedDict.get to DashOrderedDict.get_path

JosephMontoya-TRI:
Updates `DashOrderedDict.get` (which previously overrided the OrderedDict.get method) - to `get_path`, which doesn't interfere with the new pydash api.


patrickherring-TRI (approved):
LGTM, should we change all of the `get` methods to `get_path` from now on?

JosephMontoya-TRI:
Yes.  Note that `get` is still a method of OrderedDict (and therefore of DashOrderedDict), but it will not use the dot notation to get .

In fact, running `DashOrderedDict.get` with a path-like key will return None, so there's a risk of soft errors here.

PR #240, opened 2021-04-17
Loop tech part2

patrickherring-TRI:
This PR addresses changes necessary to get settings conversion to work on the BT-Lab system.
- Change abbreviations, range values and safety ranges
- Split any combined steps into separate seqs since only one control mode at a time is supported on BT-Lab system

### DIFF
--- a/beep/protocol/maccor_to_biologic_mb.py
+++ b/beep/protocol/maccor_to_biologic_mb.py
@@ -1037,7 +1037,7 @@ def convert_diagnostic_v5_segment_files():
             loop_lens[-1] += 1
 
 
-def convert_diagnostic_v5_multi_techniques():
+def convert_diagnostic_v5_multi_techniques(source_file="BioTest_000001.000"):
     def is_acceptable_goto(end_entry, step_num):
         # remove end entries going to step 70 or 94 unless
         # except when they are the next step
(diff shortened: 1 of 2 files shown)


beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -600,7 +691,7 @@ def maccor_ast_to_biologic_seqs(self, maccor_ast):
                     loop = unrolled_loop
                 else:
                     loop_seq = self._create_loop_seq(
-                        seq_num, loop_start_seq_num, num_loops
+                        seq_num, loop_start_seq_num, num_loops - 1
```
willpowelson, line 694: Just to verify this is correct, can you confirm that if I have maccor loop, and the loop value is set to `5` that the steps inbetween the `do` and the `loop` step will run precisely `5` times?
```
<Do 1>
//looped
<Loop 1>
```
(I had guessed it would run 6 times similar to biologic but had not confirmed)

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -149,7 +149,7 @@ def _convert_time(self, time_str):
             return "{:.3f}".format(total_time_sec), "s"
 
     def _proc_step_to_seq(
-        self, proc_step, step_num, seq_from_step_num, goto_lower_bound, end_step_num
+        self, proc_step, step_num, seq_from_step_num, goto_lower_bound, end_step_num, current_range='10 A'
```
willpowelson, line 152: Just to check, is the `current_range` make use of any context of the step creation here, or could we have just changed the `set_global_fields` helper for the conversion like so?:
```
def set_global_fields(seqs):
    for seq in seqs:
        seq["E range min (V)"] = "2.900"
        seq["E range max (V)"] = "4.300"
        seq["I Range"] = "10 A"
```
I'm trying to figure out if there's some context you're passing down to the converter that's specific to only _some_ steps/seqs and could not be applied to _all_ steps/seqs

willpowelson (approved):
Per our conversation, LGTM!

Issue #206, opened 2021-03-25
[mat-2490] Unittests can be organized by module


ardunn:
It is still difficult even after #150 to find where certain classes are being tested, because structuring tests are not organized by module. It can be separated by module. E.g., `test_arbin`. 

`TestStructuringUtils`

PR #241, opened 2021-04-17
Mat 2490 reorg unit test

patrickherring-TRI:
This PR reorganizes the testing for structuring to make it easier to understand where each portion of the structuring is being tested and which tests need to run to validate any changes.


ardunn:
Looks good to me, minus the merge conflict!

beep/tests/test_structure_cyclerpaths.py
```diff
@@ -0,0 +1,374 @@
+# Copyright [2020] [Toyota Research Institute]
+#
```
ardunn, line 2: I was thinking more along the lines of having these be their own files (i.e., maccor.py is tested by test_maccor_datapath.py or similar), but it doesn't really matter

ardunn (approved):
lgtm, hopefully my merge commit resolves the conflict

ardunn:
closes #206

beep/tests/test_structure_cyclerpaths.py
patrickherring-TRI: @ardunn Yeah, I guess my thought was that when you are adding a new datapath you will probably want to create the same set of tests as exist for all of the other paths. Having the tests in the same file might make that a little easier, but I could be wrong. What's your take?

ardunn: Yes that is a good point. For the time being the code looks good to me, especially since most of the datapaths only need to implement and test one method (from_file), so the file is pretty short.

PR #150, opened 2020-11-09
Structuring 2.0

ardunn:
###### Not all changes included, just the biggest ones.

- Implementing base structuring and validation class `BEEPDatapath`. Merges together `RawCyclerRun` and `ProcessedCyclerRun` with clear indication of when things are being computed (e.g., `obj.structure()` does the heavy interpolation, `obj.validate()` returns a bool for if raw data is valid or not, no running of expensive code in `__init__`s). Closes #115 
- Particular cyclers inherit this base class; this makes adding new cyclers tractable and self-contained, can be done for a new cycler with very few lines of code. Addresses problems in #115 

- Add methods:
    - `structure`: takes over the functionality of to/from_raw_cycler_run methods
    - `autostructure`: runs determine_structuring_parameters, then structures.
- Validation routines are incorporated in a `.validate()` method.
- Add more serialization support with `from_json_file` and `to_json_file` for storing and loading the raw data in/from a file (i.e., directly after ingestion) as well as all metadata and structured data; should be backwards compatible with all previously serialized `ProcessedCyclerRun` json. I.e., if you to_json_file a datapath, **ALL** of the data is stored, including paths of all source files (including eis).
- Add a `paths` attribute to keep track of original data, metadata, and processed data file locations (naively, at least). This means all the file paths related to a datapath are stored within the object. This includes things like EIS, if available
- Tests run a bit (~10%) faster

New usage looks like:
```

arbin = ArbinDatapath.from_file(my_file)

arbin.validate()
arbin.autostructure()

arbin.raw_data
>>> outputs the data equivalent of what is stored in RawCyclerRun currently


is_valid = arbin.validate()

arbin.structured_data
arbin.structured_summary
>>> outputs the data equivalents of what are stored in ProcessedCyclerRun currently

arbin.diagnostic_data
arbin.diagnostic_summary
>>> outputs the diagnostic data currently stored in ProcessedCyclerRun

arbin.to_json(some_file)
>>> writes json to disk in identical format to current beep, to maintain compatibility.

```

Adding new datapaths is now a bit less convoluted, since you can inherit from `BEEPDatapath`. For example, the complete code for implementing `ArbinDatapath` is a single `from_file` method:

```python
class ArbinDatapath(BEEPDatapath):
    @classmethod
    def from_file(cls, path, metadata_path=None):
        data = pd.read_csv(path)
        data.rename(str.lower, axis="columns", inplace=True)
        for column, dtype in ARBIN_CONFIG["data_types"].items():
            if column in data:
[...]

PR #250, opened 2021-04-27
[mat-2533] Optional saving/loading of raw_data, compression + fix test env

### LINKED ISSUE #208: Automatically serialize BEEPDatapaths to .json.gz, not json
New BEEPDatapaths are a bit bigger than older ProcessedCyclerRuns since they include complete data, including the raw data.

Running the `TestMaccorDatapath.test_get_diagnostic` and checking that file:

```
processed_diagnostic.json       249.3MB
```

Compressing it

```
processed_diagnostic.json.gz    18.6MB
```

Compressing and decompressing even large files is computationally trivial in comparison with interpolation. It seems like the default surely should be compressing these saved files.

### LINKED ISSUE #239: [Feature Request] Optional loading of `raw_data` attribute
**Problem statement**
Loading of the structured files used to be relative fast since the files contained only the interpolated dataframes and these were much smaller. Having the raw data available is very convenient but slows down the loading time and dramatically increases the amount of memory required.

**Solution suggestion**
Making the loading the of the raw data optional should solve the problem, so that unless a flag was set, the attribute would remain empty (using the `auto_load_processed` method). However, it's also desirable to make the to/from file methods simple and transparent so if a lot of additional complexity is required this might not be the best solution.

**Alternative solutions**
Another solution could be to remove the raw data attribute or turn it into a loading method that goes and looks for the original raw file. This is less desirable since it returns to the problem of locating the raw data that was used to create the structured file.

ardunn:
## Overview 

#150 made loading all datapaths very slow and memory intensive. This PR corrects this by adding an argument to `BEEPDatapath.to_json_file`, `as_legacy`, which when True does **not** save the raw data. The CLI function `process_file_list_from_json` has been changed so that the default now saves all structured files with `as_legacy=True`. All CLI-processed structured files will now be saved without raw data by default.

Also, enabled .gz/.bz2 compression in to/from_json_file.

Changes in code/usage:
- `as_legacy` argument added to `BEEPDatapath.to_json_file`. `as_legacy=True` saves the file without raw data. Default of this is to save _with raw data_.
- `omit_raw` argument added to `process_file_list_from_json`. `omit_raw=True` saves the file without raw data. Default of this is to save _without raw data_. 
- **`BEEPDatapath.to_json_file` is preferred to `monty.serialization.dumpfn`; this is so you can pass `as_legacy`, otherwise `dumpfn` will autosave the raw data.**

## Bottom line
By default, all CLI processed structured files will be saved without raw data. Individual datapaths will still save raw data by default. The to/from dict should remain as transparent as it was before.

## Benchmarking:
Example case: Loading the structured BIG_FILE as run by TestBEEPDatapath.test_get_diagnostic

Before: Loading file with raw data
```
# before:
# times: 8.452832293510436 +- 0.07645625101780719 (10 samples)
# file size: 258 MB
```

After: Loading file without raw data (as it was before #150)
```
# after:
# times: 1.6228832721710205 +- 0.012458298438177915 (10 samples)
# file size: 52 MB
```

After: Loading file without raw data (with compression):
```
# after:
# times: 1.8035454750061035 +- 0.06370021808294299 (10 samples)
# file size: 18.1 MB
```

## Notes:
- Loading files with/without raw data requires no change in `from_json_file` args. Meaning you can still call `from_json_file` or `loadfn` with the same args whether raw_data is preset or not.
- **For this speedup to work, the structured files must be saved with `as_legacy=True`. If they're not, the raw data will still load. If there are files you want to convert over, you need to load them and then save them with `as_legacy=True`. This _should_ be pretty easy.**

## Other changes
- Boilerplate import code was removed from test files, as `BIG_FILE_TESTS` was defined several places independently; in some of these, BIG_FILE_TESTS read from the `BIG_FILE_TESTS` env var, in others it was `BEEP_BIG_TESTS`. All have been changed to a single source in tests/constants.py which references `BEEP_BIG_TESTS` as is defined in the github actions workflow. 
- Some imports in test files were ordered for better readability.

## Issues closes
- Closes #239 
- Closes #208 depending on @patrickherring-TRI thoughts


ardunn:
Closes #239

ardunn:
tests should be passing, coveralls just doesn't like me

beep/structure/cli.py
```diff
@@ -137,7 +139,7 @@ def process_file_list_from_json(file_list_json, processed_dir="data-share/struct
             new_filename = add_suffix_to_filename(new_filename, "_structure")
             structured_run_loc = os.path.join(processed_dir, new_filename)
             structured_run_loc = os.path.abspath(structured_run_loc)
-            dumpfn(dp, structured_run_loc)
+            dp.to_json_file(structured_run_loc, as_legacy=omit_raw)
```
chirranjeevigopal-TRI, line 142: @ardunn what is the reason for using different args `as_legacy` and `omit_raw` for the same purpose?

chirranjeevigopal-TRI: If I want to troubleshoot some issues with the structuring and need the `raw_data` attribute again after it has been dropped, would I just have to use `from_file` method to repopulate it [since the paths to raw_data will still be stored?]?

ardunn: ## Differences in naming

> @ardunn what is the reason for using different args `as_legacy` and `omit_raw` for the same purpose?

Good observation, I will change them to be the same. Would it be ok with you to just call everything `as_legacy`? Or alternatively, `omit_raw`? If `omit_raw` I'd want to change the BEEPDatapath syntax to all be `omit_raw` instead of `legacy`. 

On second thought, that might make things really complicated, since the legacy RCR/PCR files have a separate format (no `paths` attribute, for example)... uhh let me think about this some more and put together a better solution.


## Restructuring if raw_data not saved in structured file

> If I want to troubleshoot some issues with the structuring and need the `raw_data` attribute again after it has been dropped, would I just have to use `from_file` method to repopulate it [since the paths to raw_data will still be stored?]?

I had not thought about or tested this. But yes, you'd just have to throw away your old object and restructure it from scratch using `from_file` and `structure`. Maybe I can add in some code that if the raw data path is specified but the raw data is not included, it will try to load it when using `from_json_file`. Not 100% sure this will work though. The two options then would be:

A. Keep things as they are right now. More clear code. But have to save the raw data in order to restructure it.
B. Try to autoload raw_data from paths if raw_data is not present. Less clear code. Do not have to save the raw data in order to restructure it.

@patrickherring-TRI @chirranjeevigopal-TRI do you think A or B is better?

chirranjeevigopal-TRI: > ## Differences in naming
> Good observation, I will change them to be the same. Would it be ok with you to just call everything `as_legacy`? Or alternatively, `omit_raw`? If `omit_raw` I'd want to change the BEEPDatapath syntax to all be `omit_raw` instead of `legacy`.
> 
> On second thought, that might make things really complicated, since the legacy RCR/PCR files have a separate format (no `paths` attribute, for example)... uhh let me think about this some more and put together a better solution.

Yeah looking through the base structure, I sensed that `omit_raw` might be a sub-criterion of `as_legacy` :-) but I was not entirely sure.
> 
> ## Restructuring if raw_data not saved in structured file
> > If I want to troubleshoot some issues with the structuring and need the `raw_data` attribute again after it has been dropped, would I just have to use `from_file` method to repopulate it [since the paths to raw_data will still be stored?]?
> 
> I had not thought about or tested this. But yes, you'd just have to throw away your old object and restructure it from scratch using `from_file` and `structure`. Maybe I can add in some code that if the raw data path is specified but the raw data is not included, it will try to load it when using `from_json_file`. Not 100% sure this will work though. The two options then would be:
> 
> A. Keep things as they are right now. More clear code. But have to save the raw data in order to restructure it.
> B. Try to autoload raw_data from paths if raw_data is not present. Less clear code. Do not have to save the raw data in order to restructure it.

> @patrickherring-TRI @chirranjeevigopal-TRI do you think A or B is better?
I dont think the `raw_data` has to be auto-loaded when using `from_json_file`.  I was just clarifying if the way to re-populate `raw_data` attribute for a `BeepDatapath` object after it has been dropped is to just run `from_file` with the requisite raw csv path? We could test it out after you merge.

patrickherring-TRI: @ardunn I think A is a better option since trying to remember file paths is likely to cause headaches down the line and it makes the files stateful. A better option might be a helper function that can make the path and file name lookup automatic for our data pipeline since that is well known on our system. Something like a `beep.structure.base.raw_data_reload` function or maybe a function in the utils to supply the raw data path given the name of the protocol in the metadata of the structure file.

ardunn: > Yeah looking through the base structure, I sensed that omit_raw might be a sub-criterion of as_legacy :-) but I was not entirely sure.

as_legacy has been changed to omit_raw.


> I dont think the `raw_data` has to be auto-loaded when using `from_json_file`. I was just clarifying if the way to re-populate `raw_data` attribute for a `BeepDatapath` object after it has been dropped is to just run `from_file` with the requisite raw csv path? We could test it out after you merge.

As it is right now, yes: if you didn't save the raw data, you would reload and restructure the entire file from the original raw data file.

If you _did_ save the raw data, you can load it and restructure any of the attrs you're interested in, independently. For example, you can load a structured file and restructure the diagnostic parts independently of everything else.

ardunn: > @ardunn I think A is a better option since trying to remember file paths is likely to cause headaches down the line and it makes the files stateful. A better option might be a helper function that can make the path and file name lookup automatic for our data pipeline since that is well known on our system. Something like a `beep.structure.base.raw_data_reload` function or maybe a function in the utils to supply the raw data path given the name of the protocol in the metadata of the structure file.

Ok, I will leave that for a separate PR/issue then if people feel like it is needed!

ardunn:
@patrickherring-TRI @chirranjeevigopal-TRI should be ready for merge if all looks good to you both

beep/structure/cli.py
chirranjeevigopal-TRI: Thank you for the thoughtful responses and the great work, looks good to me!

chirranjeevigopal-TRI (approved):
LGTM!

patrickherring-TRI (approved):
LGTM

PR #261, opened 2021-05-17
Add cell parameter extraction

patrickherring-TRI:
This PR adds the ability to extract cell parameters. Code developed by Patrick Asinger.


pasinger (approved):
Based on your feedback that values are coming out the same as the initial version, I think things are in a state appropriate for merging.
I made some suggestions for updating the descriptions of the primary functions in the class.

Next steps in my eyes are to 1) ensure proper function with discharge specified, 2) enable the analysis without necessarily interpolating across SOC, 3) rewrite methods to reference capacity rather than a relative value like SOC, and 4) further develop electrode blending functionality.

patrickherring-TRI:
@pasinger Agreed. I will create a ticket to work on those additional capabilities, #2 and #3 are a little fuzzy for me, but if you have time to work on them or if you have ideas about how to accomplish them let me know! I will also make subsequent PRs to add testing, clean up, etc. Also, I have not yet created the feature class from this data, but that will be another upcoming ticket.

PR #276, opened 2021-05-25
Loop advances

willpowelson:
This PR is a first attempt at calculating cycle number in output data based on the protocol used to generate it.

For @patrickherring-TRI 
The basic scheme (which handles a deceptively large number of edge cases)

- Preprocess the steps based on how you will split them into techniques, now each technique can only have loops one deep
- For each technique we create 
  - a mapping of transitions between physical steps (e.g. constant current) that advance cycle and how much. **NOTE** We do not handle cycle advancement via GOTOs currently, but they are not important to our current protocols and can be added to this scheme later without any friction.
  - the number of cycle advancement to perform when a technique loops
  - the number of cycle advancements to perform when a technique ends

there is some additional data included primarily for debugging purposes, but it is not important.

We then combine this step transition data with the `step_num` and `seq_num` mapping to create a map of number of cycle advancements on a `seq_num` transition.

Concretely this mapping might look something like:
```
{
  (5, 6): 1, when seq num changes from 5 to 6, advance cycle num 1
  (9, 11): 1, when seq num changes from 9 to 11, advance cycle num 1
   ...
}
```
This format is not JSON serializable so I've created a helper class to assist with serialization/deserialization.

@PatrickMoore-TRI 
Now I'll describe the part of this that relates to the pipeline. For some context, this code is pretty rough and ready because we're not actually sure if it will work, and if it does we're going to have to rewrite most of the main converter class but if this works I've tried to structure things so it the cleanup won't impact the pipeline or anyone downstream of this file.

You should only need to worry about the code in:
- the `CycleTransitionRules` class
- the `CycleTransitionRulesSerializer` class
- the `add_cycle_nums_to_csvs` function

If this scheme works, from now on all  data for a run will be split by its generating "technique". The code here determines the technique (note that not all techniques generate data, so the numbers may not be contiguous e.g. 1, 2 4). We will calculate ahead of time rules for incrementing the cycle number in data based on the protocol that generated the data. Once the data comes in we will use these rules to add a cycle number column to the csv data.

Because cycle number is global to a run, we will need to load the CSV files sequentially. Determining the cycle number for data from a given technique requires knowing the cycle number at the end of the previous technique.

Many runs will share a JSON encoding of cycle advancement rules.

I've created a sample function of what this process might look like with `add_cycle_nums_to_csvs`, here's what a call for it might look like for diagnosticV5.000
```
add_cycle_nums_to_csvs(
    [
      os.path.join(MY_DIR, "protocol1_a_tech_1.csv")],
      os.path.join(MY_DIR, "protocol1_a_tech_2.csv")],
    ],
    [
      os.path.join(MY_DIR, "protocol1_tech_1_rules.json")],
      os.path.join(MY_DIR, "protocol1_tech_2_rules.json")],
    ],
    [
      os.path.join(MY_DIR, "protocol1_a_tech_1_out.csv")],
      os.path.join(MY_DIR, "protocol1_a_tech_2_out.csv")],
    ],
)
```
I'm not quite sure how to fit this into pipeline, so your help is greatly appreciated there. You may, perhaps need to know ahead of time which technique numbers to be looking for in the data (which we can generate trivially).

### DIFF
--- a/beep/protocol/maccor_to_biologic_mb.py
+++ b/beep/protocol/maccor_to_biologic_mb.py
@@ -151,8 +151,14 @@ class MaccorToBiologicMb:
             return "{:.3f}".format(total_time_sec), "s"
 
     def _proc_step_to_seq(
-        self, proc_step, step_num, seq_from_step_num, goto_lower_bound, end_step_num,
-            current_range='10 A', split_step=None
+        self,
+        proc_step,
+        step_num,
+        seq_from_step_num,
+        goto_lower_bound,
+        end_step_num,
+        current_range="10 A",
+        split_step=None,
     ):
         """
         converts steps that are not related to control flow to sequence dicts
@@ -398,9 +404,7 @@ class MaccorToBiologicMb:
 
         return new_seq
 
-    def _split_combined_step(
-        self, step, step_num
-    ):
+    def _split_combined_step(self, step, step_num):
         """
         converts steps that have a combined control mode (CCCV) into separated steps that are compatible with BT-Lab
         """
@@ -408,12 +412,17 @@ class MaccorToBiologicMb:
         step_part2 = clone_deep_with(step)
 
         if isinstance(get(step_part1, "Ends.EndEntry"), list):
-            indx = find_index(get(step_part1, "Ends.EndEntry"), lambda x: x["EndType"] == "Current")
+            indx = find_index(
+                get(step_part1, "Ends.EndEntry"), lambda x: x["EndType"] == "Current"
+            )
             path = "Ends.EndEntry.{}".format(indx)
         else:
             path = "Ends.EndEntry"
 
-        if step_part1["StepMode"] == "Current" and "Voltage" in step_part1["Limits"].keys():
+        if (
+            step_part1["StepMode"] == "Current"
+            and "Voltage" in step_part1["Limits"].keys()
+        ):
             if step_part1["StepType"] == "Charge":
                 set_(step_part1, path + ".Oper", ">=")
             elif step_part1["StepType"] == "Dischrge":
@@ -429,7 +438,10 @@ class MaccorToBiologicMb:
             step_part2["StepValue"] = step["Limits"]["Voltage"]
             step_part2["Limits"] = None
 
-        elif step_part1["StepMode"] == "Voltage" and "Current" in step_part1["Limits"].keys():
+        elif (
+            step_part1["StepMode"] == "Voltage"
+            and "Current" in step_part1["Limits"].keys()
+        ):
             if step_part1["StepType"] == "Charge":
                 set_(step_part1, path + ".Oper", ">=")
             elif step_part1["StepType"] == "Dischrge":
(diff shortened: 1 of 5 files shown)


moorepatrick (approved):
LGTM

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -1271,3 +1645,61 @@ def set_global_fields(seqs):
...
+            cycle_transition_rules = serializer.parse_json(data)
+
+        df = pd.read_csv(csv_fp, sep=";")
+
+        cycle_num += cycle_transition_rules.adv_cycle_on_start
+
+        prev_seq_num = int(df.iloc[0]["Ns"])
+        prev_loop_num = int(df.iloc[0]["Loop"])
+        cycle_nums = []
+        for _, row in df.iterrows():
```
patrickherring-TRI, line 1665: Any benchmark on how long this takes to run?

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -1271,3 +1645,61 @@ def set_global_fields(seqs):
     with open(fp, "wb") as f:
         f.write(file_str.encode("ISO-8859-1"))
         print("created {}".format(fp))
+
+
+def add_cycle_nums_to_csvs(csv_and_transition_rule_file_paths):
```
patrickherring-TRI, line 1650: Is there a reason to keep this as a single argument instead of splitting out each of the components?

patrickherring-TRI (commented):
Just a few questions

beep/protocol/maccor_to_biologic_mb.py
willpowelson: Not particularly, we can change it if you prefer.

beep/protocol/maccor_to_biologic_mb.py
willpowelson: System Specs:
Macbook Pro
16 GB 2400 MHz DDR4
2.6 GHz 6-Core Intel Core i7
SSD Hard Drive

Processing a 1.7gb file takes 10 minutes.

Pretty atrocious in absolute terms. I haven't done much CSV processing in Python so I'm not entirely sure what standard performance for the task is but I imagine once we verify the logic is correct we can speed it up by an order of magnitude.

beep/protocol/maccor_to_biologic_mb.py
patrickherring-TRI: If it's not too much of a pain, that might help clarify how to use the function. Alternatively, a small doc string that has an example input so that I can see what kinds of arguments are needed (absolute path, list of paths, etc.)

patrickherring-TRI (changes requested):
If you could just add a comment or something about what the input for the function should look like

beep/protocol/maccor_to_biologic_mb.py
willpowelson: I split the arg into 3, added some doc strings. Please let me know if they're unclear.

patrickherring-TRI (approved):
LGTM

PR #277, opened 2021-05-26
Add the battery archive class to the init

patrickherring-TRI:
Adding battery archive class to the structure imports


ardunn (approved):
LGTM

PR #279, opened 2021-05-27
Add epoch time as a column in throughput calculations

patrickherring-TRI:
This PR adds a column "epoch_time" that is the timestamp in seconds of the cycle metric that is being calculated. This should make it possible to use epoch time in the threshold calculations

### DIFF
--- a/beep/dataset.py
+++ b/beep/dataset.py
@@ -344,6 +344,7 @@ class BeepDataset(MSONable):
 def get_threshold_targets(dataset_diagnostic_properties,
                           cycle_type="rpt_1C",
                           metric="discharge_energy",
+                          x_axis_types=None,
                           threshold=0.80,
                           filter_kinks=None,
                           extrapolate_threshold=True):
@@ -361,6 +362,7 @@ def get_threshold_targets(dataset_diagnostic_properties,
             DiagnosticProperties feature
         cycle_type (str): Type of diagnostic cycle being used to measure the fractional metric
         metric (str): The metric being used for fractional capacity
+        x_axis_types (list): List of column names to use for x_axis interpolation (distance to threshold)
         threshold (float): Value for the fractional metric to be considered above or below threshold
         filter_kinks (float): If set, cutoff value for the second derivative of the fractional metric (cells with an
             abrupt change in degradation rate might have something else going on). Typical value might be 0.04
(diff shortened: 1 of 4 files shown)


moorepatrick (commented):
I'd like to let someone else who uses this code give the approval, but is the x_axis calculations commit related?

patrickherring-TRI:
@moorepatrick Yes, the updated x_axis calculation will use the time basis that is being added to the features.

pasinger (approved):
LGTM!

pasinger (approved):
LGTM

beep/dataset.py
```diff
@@ -400,39 +408,45 @@ def get_threshold_targets(dataset_diagnostic_properties,
         elif run_target_df['fractional_metric'].min() > threshold and extrapolate_threshold:
             fill_value = "extrapolate"
             bounds_error = False
-            x_throughput_linspace = np.linspace(x_throughput_axis.min(), 2 * x_throughput_axis.max(), num=1000)
-            x_cycle_linspace = np.linspace(x_cycle_axis.min(), 2 * x_cycle_axis.max(), num=1000)
+            x_linspaces = []
+            for axis in x_axes:
+                x_linspaces.append(np.linspace(axis.min(), 2 * axis.max(), num=1000))
```
pasinger, line 413: Why does the linspace upper bound go to 2*axis.max() instead of just axis.max()?

pasinger: Ah I see now that this is for cases where the fractional_metric has not yet crossed the threshold. I see why there should be a factor greater than 1, but a value of 2 feels arbitrary.
If a test is very early in its lifetime and extrapolation is on, then 2*axis.max() may still not provide a value that will go below the threshold.
Could it be made more specific such as by making the upper bound the predicted linear extrapolation:
x_thresh_extrap = (threshold - y1)*(x2-x1)/(y2-y1)+x1, where
 y1 = y_interpolation_axis.iloc[-2]
y2 = y_interpolation_axis.iloc[-1]
x1 = axis[-2]
x2 = axis[-1]
We could make the extrapolation go a little beyond what's exactly necessary by subtracting another 0.1 from threshold, for example.

beep/dataset.py
```diff
@@ -400,39 +408,45 @@ def get_threshold_targets(dataset_diagnostic_properties,
...
-            x_cycle_linspace = np.linspace(x_cycle_axis.min(), x_cycle_axis.max(), num=1000)
+            x_linspaces = []
+            for axis in x_axes:
+                x_linspaces.append(np.linspace(axis.min(), axis.max(), num=1000))
 
-        f_throughput = interpolate.interp1d(x_throughput_axis, y_interpolation_axis, kind='linear',
-                                            bounds_error=bounds_error, fill_value=fill_value)
-        f_cycle = interpolate.interp1d(x_cycle_axis, y_interpolation_axis, kind='linear', bounds_error=bounds_error,
-                                       fill_value=fill_value)
+        f_axis = []
```
pasinger, line 421: Can we move the this initialization of f_axis to below that of x_linspaces, then combine the below code (second for loop over x_axes) to with the first for loop over x_axes?

pasinger (commented):
This looks like a nice update to enable the retrieval of more types of measured values at the threshold value. I made a few suggestions, but it looks to me like it will work just as well as it was previously without making the suggested changes.

PR #286, opened 2021-06-02
Changes 2021

cx26:


beep/features/featurizer_helpers.py
```diff
@@ -397,174 +397,124 @@ def get_hppc_ocv(processed_cycler_run, diag_pos):
...
+    if r_type[2] == 'd':
+        # one hour long rest 
+        step_ocv = 0
+        step_cur = 1
+    
+    # if there is no dataframe for ocv or step cur, it means this step is skipped, so return None directly 
+    try:
+        chosen_step_ocv = chosen[(chosen.step_index_counter == counters[step_ocv][soc])]
+        chosen_step_cur = chosen[chosen.step_index_counter == counters[step_cur][soc]]
+    except:
```
patrickherring-TRI, line 439: Its usually best practice to catch a specific type of exception that you expect. For instance, here you might be expecting a `KeyError` or something similar, based on the comment above. btw, thank you for adding the comment above since it makes it much clearer what is going on.

So it would look like
except KeyError:

beep/tests/test_features.py
```diff
@@ -309,7 +309,7 @@ def test_HPPCResistanceVoltageFeatures_class(self):
             self.assertEqual(featurizer.X.shape[1], 76)
             self.assertListEqual(
                 [featurizer.X.columns[0], featurizer.X.columns[-1]],
-                ["ohmic_r_d0", "D_8"],
+                ["r_c_0s_00", "D_8"],
```
patrickherring-TRI, line 312: It would be helpful to add some testing values here to verify that the calculation is being done correctly.

patrickherring-TRI (changes requested):
Just some minor changes and fix up some linting errors:
```
2       E225 missing whitespace around operator
2       E251 unexpected spaces around keyword / parameter equals
1       E303 too many blank lines (2)
1       E722 do not use bare 'except'
```

beep/features/featurizer_helpers.py
cx26: Thank you Patrick! I added in except "IndexError"

beep/tests/test_features.py
cx26: Added in some numerical values, and the calculation is the same with my notebook, but I encountered this error "ubuntu-latest_3_7 can't be loaded in"

patrickherring-TRI (approved):
LGTM!

PR #319, opened 2021-07-07
Working version of new interpolation

patrickherring-TRI:
This PR changes the interpolation of the diagnostic cycles so that cv steps are interpolated on time instead of voltage. This is important so that enough points are included for feature extraction.

### DIFF
--- a/beep/structure/base.py
+++ b/beep/structure/base.py
@@ -816,7 +816,7 @@ class BEEPDatapath(abc.ABC, MSONable):
 
     # equivalent of get_interpolated_diagnostic_cycles
     def interpolate_diagnostic_cycles(
-            self, diagnostic_available, resolution=1000, v_resolution=0.0005
+            self, diagnostic_available, resolution=1000, v_resolution=0.0005, v_delta_min=0.001
     ):
         """
         Interpolates data according to location and type of diagnostic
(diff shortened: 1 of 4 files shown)


MuratAykol-TRI (approved):
Looks good!

d-cogswell (commented):
I ran some tests and can confirm that CV steps which didn't have data before are now being interpolated. However I noticed a couple issues:
-There is a lot of constant voltage data stored in CCCV steps, and these changes do not affect CCCV interpolation. The CV data is for these still largely missing. Maybe CCCV also needs a dt condition for interpolation?
-There are significantly more interpolated data points than in the original raw data, which might not be great for performance.

What do you think about adding the first and last data point for every step_index? That might be enough for me to run my initial analysis. Right now many of the steps don't have any data.

patrickherring-TRI:
@d-cogswell My concern is that many of the existing featurization methods rely on a specific number of data points that are evenly spaced throughout the voltage window (its the primary reason for interpolating). So if we add a dt condition to the interpolation we will effectively break most of those functions. Same would be true of adding the first and last data point. 

What I can do instead is add a summary statistic that is computed for the diagnostic steps that are CV or CCCV. If you have some code that is computing the relevant information from the raw data, we can just call that function during the structuring method and include the values that it computes. That way the method has the cycle type information, and access to the raw data at the same time.

d-cogswell (approved):

PR #320, opened 2021-07-12
Working version of intracell features

patrickherring-TRI:
This PR adds features derived from the intracell analysis. Curve fitting using half cell data is used to determine the parameters of the full cell using the most recent dQdV data.
- Adds IntracellCycles feature which has rows which correspond to the LLI, LAM_NE, LAM_PE, etc. for each diagnostic in the life of the cell
- Add IntracellFeatures feature which has a single row with the LLI, LAM_NE, LAM_PE, etc. from the first and second diagnostic cycle broadcast together


MuratAykol-TRI (approved):
Overall looks good to me! A minor point that we may maybe tag as future cleanup is that `IntracellCycles` and `IntracellFeatures` have significant code overlap, which could be mitigated by parameterizing:
- the limits of the loop on line 121
- and whether resulting `degradation_df` is returned directly or the few additional steps are taken to get and return `feature_df`

patrickherring-TRI:
@MuratAykol-TRI Yes, I think we should address the code overlap in a future PR. The complication is that the class is supposed to output one feature object per class, so to have two feature objects being written might require us to rewrite the class structure

PR #327, opened 2021-07-13
CV summary stats

d-cogswell:
Adds cv_time and cv_current as summary stats to the regular cycles and diagnostic cycles. cv_time is the length of the CV segment of charge (useful for CV with current limit), and cv_current is the current at the end of the segment (useful for CV with time limit).

Also changes test_time from float32 to float64 in order to handle large times without decimal truncation.

### DIFF
--- a/beep/structure/base.py
+++ b/beep/structure/base.py
@@ -796,6 +796,18 @@ class BEEPDatapath(abc.ABC, MSONable):
         summary["paused"] = self.raw_data.groupby("cycle_index").apply(
             get_max_paused_over_threshold)
 
+        # Add CV_time and CV_current summary stats
+        CV_time = []
+        CV_current = []
+        for cycle in summary.cycle_index:
+            raw_cycle = self.raw_data.loc[self.raw_data.cycle_index == cycle]
+            charge = raw_cycle.loc[raw_cycle.current > 0]
+            CV = get_CV_segment_from_charge(charge)
+            CV_time.append(CV.test_time.iat[-1] - CV.test_time.iat[0])
+            CV_current.append(CV.current.iat[-1])
+        summary["CV_time"] = CV_time
+        summary["CV_current"] = CV_current
+
         summary = self._cast_dtypes(summary, "summary")
 
         last_voltage = self.raw_data.loc[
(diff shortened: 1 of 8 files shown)


d-cogswell:
Where is the best place to add a test of the CV_time and CV_current stats?

beep/structure/base.py
```diff
@@ -1385,3 +1412,62 @@ def get_max_paused_over_threshold(group, paused_threshold=3600):
...
+
+    """
+    # Compute dI and dV
+    dI = np.diff(charge.current)
+    dV = np.diff(charge.voltage)
+    dt = np.diff(charge.test_time)
+
+    # Find the first index where dt>1 and abs(dV/dt)<tol and abs(dI/dt)>tol
+    i = 0
+    while i < len(dV) and not (dt[i] > 1 and abs(dV[i]/dt[i]) < 5.e-5 and abs(dI[i]/dt[i]) > 1.e-4):
```
patrickherring-TRI, line 1436: Also, is there a better way of evaluating this condition or narrowing down the number of points this is evaluated on? There are two nested loops here that give me pause. Can we try this code out on a full size file to check the compute time?

patrickherring-TRI (changes requested):
It would be helpful to add some testing here. Both unit tests for the functions and the structuring class to make sure that the right changes are reflected (make sure that the correct columns are now in the structured output). Also if you have run some tests on full scale files, it would be nice to verify that the additional compute time is reasonable.

beep/structure/base.py
d-cogswell: I used the while loop to reduce the number of evaluated points. When the beginning of CV is reached, the loop breaks and no more points are evaluated. Would it be faster to avoid the while loop but evaluate every point? I can measure some timings.

I think the nesting might be necessary. For every cycle, you have loop over the charge points and find the start of CV.

d-cogswell: Evaluating get_CV_segment_from_charge() on ~900 cycles of raw data takes ~9 seconds on my workstation. Most of that time is filtering each cycle for charge, but replacing the inner while loop with vectorized numpy calls saves ~1sec. Should I make this change?

patrickherring-TRI: I guess if its only a small improvement we can leave as is for now, but it might be nice to include the numpy call as a comment in case we need a speed improvement.

d-cogswell: The numpy approach had problems with boundary cases (empty arrays and data with no CV), so let's stick with the loop for now.

patrickherring-TRI (approved):
LGTM! Thanks Dan!

PR #328, opened 2021-07-16
MAT-2686

danielschweigert-TRI:
Changes to biologic data file parsing to accommodate new file layout.


moorepatrick (approved):
LGTM

beep/conversion_schemas/biologic_conversion.yaml
```diff
@@ -6,7 +6,7 @@ data_columns:
   cycle number:
```
patrickherring-TRI, line 6: @willpowelson @moorepatrick Is the cycle number where the cycle index being stored?

willpowelson: no, cycle index is stored under the header `cycle_index` which we add to the files in the pipeline.

Cycle number is calculated by Biologic based on charge/discharge alternance and is not what we want.

willpowelson: see https://github.com/TRI-AMDD/beep-sync/pull/17#issuecomment-870810296
for the discussion where we debugged this in BEEP-sync

patrickherring-TRI, line 6: @danielschweigert-TRI Can you change this to cycle_index?

patrickherring-TRI (changes requested):
I think that cycle_index should be a column already in the data, could you use that value instead?

beep/conversion_schemas/biologic_conversion.yaml
danielschweigert-TRI: @patrickherring-TRI I don't see a `cycle_index` among the column names. The column names header is:
```
mode;ox;error;time/s;Ecell/V;dq/mA.h;I/mA;(Q-Qo)/mA.h;x;control/V;control/mA;control changes;dQ/mA.h;cycle number;DQ/mA.h;Rapp/Ohm;control/°C;T/°C;rotation rate/rpm;Ns changes;freq/Hz;|Ewe|/V;|I|/A;Phase(Z)/deg;|Z|/Ohm;Re(Z)/Ohm;-Im(Z)/Ohm;I Range;Q charge/mA.h;Q discharge/mA.h;Tech Num;End buf;counter inc.;R/Ohm;P/W;Energy/W.h;Analog OUT/V;Cs-2/µF-2;Re(Y)/Ohm-1;Im(Y)/Ohm-1;|Y|/Ohm-1;Phase(Y)/deg;Energy charge/W.h;Energy discharge/W.h;Capacitance charge/µF;Capacitance discharge/µF;Ns;dI/dt/mA/s;Delta(mass)/g;Phase(I)/rad;Rcmp/Ohm;Cs/µF;sin ampl/V;Conductivity/S.cm-1;Cp/µF;Cp-2/µF-2;Efficiency/%;(Q-Qo)/C;dQ/C;cycle time/s;charge time/s;discharge time/s;d(Q-Qo)/dE/mA.h/V;Capacity/mA.h;<EcellX>/V;Temperature/°C/°c;z cycle;NbIterInstr;Instr;Loop
```

patrickherring-TRI:
You might have the wrong file... @moorepatrick did you send the file resulting from concatenation?

beep/conversion_schemas/biologic_conversion.yaml
patrickherring-TRI: Those look like the column names pre concatenation... I can look at this later and regenerate the file if necessary

danielschweigert-TRI: My fault; I picked the wrong one which is similarly named and one of the pre-concatenated ones.

However, the correct file file does not appear to have a corresponding meta-data file (with `.mpl` extension) in the bucket.

patrickherring-TRI: It should be the same metadata file as the other file... You can just rename the metadata file to match and we will fix that in the syncer

patrickherring-TRI (approved):
LGTM!

PR #335, opened 2021-07-23
Change versions of boto and update tests

patrickherring-TRI:
This PR updates the versions of boto to resolve some dependency issues. It also updates the featurization test from raw data. It appears that the switch of test_time and test (sec) from float32 to float64 changes the resistance values in the feature.


pasinger (approved):
Thanks for addressing these changes, PK. Looks good to me

PR #342, opened 2021-07-28
MAT-2798 Update Maccor to Biologic Conversion

willpowelson:
My apologies to your day.

This PR primarily moves our conversion strategy away from using loop unrolling to create nested loops and instead breaks the protocol into techniques.

Several learnings from doing a variety of previous conversions have been incorporated:

- in maccor it's common to have EndEntry tags for max and min voltage or current. Simply set any of
  - max_voltage_v
  - min_voltage_v
  - max_current_a
  - max_current_a

  and all such end entries will be filtered from the incoming steps and the appropriate fields will be set in all of the outgoing seqs
- want to alter one or more steps with some custom logic? append a function with the signature `(step, step_num) -> step` to the `step_mappers` field of your instance of `MaccorToBiologic` and it will run on all the steps.
- mappers can be cumbersome to write if you just want to remove certain EndEntry tags; you can check the `_create_step_end_entry_filter` helper method to make this easier.
- If you want to alter the output of certain seqs, simply add a mapper to the `seq_mappers` field of your instance of `MaccorToBiologic` and it will run on all the seqs. (we will probably want to do this regularly with the `I Range` field)

Sample Code
```python
converter = MaccorToBiologic()
coverter.max_voltage_v = 4.4
coverter.min_voltage_v = 2.2
def set_i_range(tech_num, seq, idx):
  seq_copy = copy.deepcopy(seq)
  seq_copy["I Range"] = "10 A"
  return seq_copy
converter.seq_mappers.append(set_i_range)

conveter.convert(
  target_maccor_file_filepath,
  output_dir,
  output_filename,
)
```

A few more involved tests will need to be written once we've actually tested the output a bit more.


beep/protocol/procedure_templates/a_formation06162921_Xiao.000
```diff
@@ -0,0 +1,611 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
patrickherring-TRI, line 1: @willpowelson Could you remove this file from the PR? I think it might be better to keep it out of the repo since it involves ongoing work and might be confusing... :)

willpowelson: @patrickherring-TRI  it's gone!

patrickherring-TRI (approved):
LGTM

PR #349, opened 2021-08-03
add test and validation for non-overlapping intracell criteria

JosephMontoya-TRI:
* Adds a condition to `IntracellFeatures.validate_data` to test for empty eol diagnostic cycles.
* Adds an associated test
* Renames constants in `IntracellAnalysis` to be accessible at the class, rather than instance level of the object


patrickherring-TRI (approved):
LGTM! Thanks for your help with this!

PR #360, opened 2021-08-17
Update converter

patrickherring-TRI:
This PR makes changes to the conversion to fix some minor bugs.


beep/tests/test_maccor_to_biologic_mb.py
```diff
@@ -409,6 +415,83 @@ def test_discharge_current_step_conversion(self):
...
+        converter.min_voltage_v = 0
+        converter.max_voltage_v = 4.45
+
+        converter.convert("/Users/patrickherring/Code/beep/beep/protocol/procedure_templates/formation_061621.000",
+                          "/Users/patrickherring/Code/beep/beep/protocol/biologic_templates", "formation_061621")
+
+        print(converter._mps_header_template[225:259])
+        self.assertEqual(converter._mps_header_template[225:260], '\tEcell min = {}\r\n\tEcell max = {}\r\n\t')
+
+    def test_conversion_with_updated(self):
```
willpowelson, line 436: This test feels a little obtuse to me, it looks like we're doing a lot of stuff to essentially test whether the headers are being formatted correctly when the global voltage limits are set. Is that right or does this test something else?

If not consider just loading up any diagnostic we plan to hang onto, convert it, and check the header. No need to even set the I Range

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -510,8 +511,17 @@ def _diagnostic_file_str_from_technique_partitions_post_conversion(self, techniq
         assert len(technique_partitions_post_conversion) > 0
         last_tech = technique_partitions_post_conversion[-1]
         num_techniques = last_tech.technique_num + 1 if last_tech.tech_does_loop else last_tech.technique_num
+        # Insert safety limits into the header for the file output
+        if self.min_voltage_v is None:
+            safety_min_v = "0 V"
+        else:
+            safety_min_v = str(self.min_voltage_v) + " V"
+        if self.max_voltage_v is None:
+            safety_max_v = "4.45 V"
```
willpowelson, line 520: If we want defaults, why not just set them in the constructor and let the class throw if the user sets them to `None`?

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -510,8 +511,17 @@ def _diagnostic_file_str_from_technique_partitions_post_conversion(self, techniq
         assert len(technique_partitions_post_conversion) > 0
         last_tech = technique_partitions_post_conversion[-1]
         num_techniques = last_tech.technique_num + 1 if last_tech.tech_does_loop else last_tech.technique_num
+        # Insert safety limits into the header for the file output
+        if self.min_voltage_v is None:
```
willpowelson, line 515: If we want defaults, why not just set them in the constructor and let the class throw if the user sets them to `None`?

willpowelson (commented):
Details in the comments, we may want a few changes here but overall looking good (requires linting)

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -454,8 +454,8 @@ def _create_loop_seq(self, seq_num, seq_num_to_loop_to, num_loops):
         loop_seq = self._blank_seq.copy()
         loop_seq["Ns"] = seq_num
         loop_seq["ctrl_type"] = "Loop"
-        loop_seq["ctrl_repeat"] = num_loops
-        loop_seq["ctrl_seq"] = seq_num_to_loop_to
+        loop_seq["ctrl_repeat"] = num_loops - 1
+        loop_seq["ctrl_seq"] = seq_num_to_loop_to -1
```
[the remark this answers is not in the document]
patrickherring-TRI: The mapping seems to be off by one, there might be other places to make this change, but this was the clearest place that I could find.

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -510,8 +511,17 @@ def _diagnostic_file_str_from_technique_partitions_post_conversion(self, techniq
         assert len(technique_partitions_post_conversion) > 0
         last_tech = technique_partitions_post_conversion[-1]
         num_techniques = last_tech.technique_num + 1 if last_tech.tech_does_loop else last_tech.technique_num
+        # Insert safety limits into the header for the file output
+        if self.min_voltage_v is None:
+            safety_min_v = "0 V"
+        else:
+            safety_min_v = str(self.min_voltage_v) + " V"
```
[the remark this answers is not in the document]
patrickherring-TRI: Done

beep/tests/test_maccor_to_biologic_mb.py
patrickherring-TRI: I will write this into the other test and delete this test

beep/protocol/maccor_to_biologic_mb.py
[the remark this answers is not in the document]
willpowelson: ok let's put a TODO to fix the seq mapping off by one error

beep/protocol/maccor_to_biologic_mb.py
patrickherring-TRI: Done

beep/protocol/maccor_to_biologic_mb.py
patrickherring-TRI: Done

beep/protocol/maccor_to_biologic_mb.py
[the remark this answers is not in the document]
patrickherring-TRI: Done

beep/tests/test_maccor_to_biologic_mb.py
willpowelson: Were you going to delete this?

patrickherring-TRI: I deleted the other test that was just changing the header, this is the test that can run with the existing files

willpowelson: I still don't quite understand why this test isn't
```python
converter = MaccorToBiologicMb()
converter.min_voltage_v = 2.0
converter.max_voltage_v = 4.4
def set_i_range(tech_num, seq, idx):
  seq_copy = copy.deepcopy(seq)
  seq_copy["I Range"] = "100 mA"
  return seq_copy

converter.convert(
  os.path.join(scratch_dir, "BioTest_000001.000"),
  TEST_FILE_DIR,
  "BioTest_000001"
 )
 
 self.assertEqual(converter._mps_header_template[225:260], '\tEcell min = {}\r\n\tEcell max = {}\r\n\t')
 ```
could you explain what all the procedure generation code is getting us?

Comment (patrickherring-TRI) [reply beep/tests/test_maccor_to_biologic_mb.py:418]: The generation code here is just to create a version of the protocol that can be tested with cells that are currently connected to the cycler. As such, I wanted to expose the parameters that we are using so that modifications can be made easily.

Comment (willpowelson) [reply beep/tests/test_maccor_to_biologic_mb.py:418]: Ok so if I understand correctly this test is to make sure that with the latest up-to-date generated maccor protocols the converter works.

I think that's fine but wonky enough that we should add a comment because it's very non-obvious looking at the test.

**Review by willpowelson — APPROVED** [review]
LGTM

File: beep/tests/test_maccor_to_biologic_mb.py (hunk 1 above)
Comment (patrickherring-TRI) [reply beep/tests/test_maccor_to_biologic_mb.py:418]: Oh, the test that you have in the comment is actually already gone.

Comment (willpowelson) [reply beep/tests/test_maccor_to_biologic_mb.py:418]: `test_conversion_with_updated` is still appearing to me in the diff

File: beep/tests/test_maccor_to_biologic_mb.py [hunk 6]
```diff
@@ -409,6 +415,81 @@ def test_discharge_current_step_conversion(self):
...
+            diagnostic_params = diag_params_df[
+                diag_params_df["diagnostic_parameter_set"]
+                == protocol_params["diagnostic_parameter_set"]
+                ].squeeze()
+
+            procedure = Procedure.generate_procedure_regcyclev3(index, protocol_params)
+            procedure.generate_procedure_diagcyclev3(
+                protocol_params["capacity_nominal"], diagnostic_params
+            )
+            procedure.set_skip_to_end_diagnostic(4.4, 2.0, step_key="070", new_step_key="095")
```
willpowelson, line 459: this is the section that confused me

beep/tests/test_maccor_to_biologic_mb.py
patrickherring-TRI: yes, that was the test I wanted to keep, but the code you have in the comment should not be there anymore

willpowelson: I think there's some confusion over what code is confusing me, lines 423-461 of `test_maccor_to_biologic.py` in the latest commit of this branch `043d630388a997b24e3b3cf1e3e337f634d5e6f3`. I left a comment elsewhere highlighting the confusing code.

patrickherring-TRI: Got it.

beep/tests/test_maccor_to_biologic_mb.py
patrickherring-TRI: Got it. Added some comments to at least highlight what is going on

beep/protocol/maccor_to_biologic_mb.py
```diff
@@ -986,36 +994,41 @@ def _convert_steps_to_seqs(self, steps, step_num_offset, end_step_num):
...
                     raise Exception("Unclosed loop at {}".format(step_num))
 
                 assert len(seq_nums_by_step_num[step_num]) == 1
                 seq_num = seq_nums_by_step_num[step_num][0]
+                num_loops = max(
+                    # the way maccor and biologic count loops varies by 1
+                    int(get(step, "Ends.EndEntry.Value")) - 1,
+                    0,
+                )
+                seq_num_to_loop_to = seq_nums_by_step_num[prev_loop_open_step_num][0]
```
willpowelson, line 1021: @patrickherring-TRI I've moved any finagling with where to loop to or how many loops here because to make the code easier to follow. This should fix the goto looping bug we discussed.

PR #361, opened 2021-08-18
Resolve bugs in IntracellAnalysis.py

pasinger:
Resolves a bug in IntracellAnalysis.py. NaN values were propagating into the error metric for the optimizer, which led to the optimizer triggering an error. Now, the NaN values are dropped to avoid this error.


patrickherring-TRI (approved):
LGTM

PR #370, opened 2021-08-27
Extend intracell threshold

pasinger:
Change the threshold for conducting the intracell analysis from 0.8 to 0.7, so that we may be able to include at least one cycle that goes below 0.8. This way, we can interpolate the final states at 0.8 rather than extrapolate. Before this change, we would be guaranteed to not include the diagnostic with fractional metric below 0.8.


patrickherring-TRI (approved):
LGTM

PR #386, opened 2021-09-21
Decrease the tolerance for more accurate extraction of CV from CCCV

d-cogswell:
CV was occasionally not being extracted correctly for some cycles (see for example 'PreDiag_000447'). Part of the CC segment is sometimes included as CV. Decreasing the tolerance settings seems to fix the issue.


patrickherring-TRI (approved):
LGTM!

PR #403, opened 2021-09-29
Fix error in converting formation files

patrickherring-TRI:
This fixes a bug for conversion of files where the goto in the maccor file points to the end step which has been removed. It is a combination of an off-by-one error and a typo in the variable name.


willpowelson (approved):
LGTM

PR #404, opened 2021-09-30
test for goto end bug

willpowelson:
This PR adds a test for[ this pull request ](https://github.com/TRI-AMDD/beep/pull/403) where the converter was crashing when the protocol contained a step that could GOTO the end step, and that step did not immediately precede the end step.

This PR includes a minimal example protocol: 2 physical steps, 1 end step, where the first step GOTOs the end step and a test that converts it.


patrickherring-TRI (approved):
LGTM

PR #405, opened 2021-09-30
Change the order to address the case of the final step goto pointing …

patrickherring-TRI:
…to the end step


patrickherring-TRI:
@willpowelson Per our conversation, this might cause more problems than it fixes. Will close for now and reopen if and when it becomes an issue

PR #406, opened 2021-10-01
Fix CLI tests outputs

ardunn:
Previously, stacktraces could not be captured in click+pytest for some reason, this just does not capture the stacktrace by preventing click CliRunner from internally catching exceptions


patrickherring-TRI (approved):
LGTM

PR #418, opened 2021-10-08
Fix matplotlib requirement install issue

ardunn:
Package would not install/run as-is due to matplotlib needing to be imported in some modules, and it not being listed under requirements.

Also, removed docopt as it is not needed.


ardunn:
@mi-barry
@patrickherring-TRI

patrickherring-TRI (approved):
LGTM

PR #426, opened 2021-10-13
Extending an existing test for reinvoking

patrickherring-TRI:
Errors were occurring due to changes to the hyperparameter dictionary done in place. This test makes sure that this is no longer an error.


amalie-trewartha:
Looks good!

amalie-trewartha (approved):

PR #465, opened 2021-11-16
Add python 3.9 for integration tests

shijingsun-TRI:
Add python 3.9 for integration tests

PR #478, opened 2021-12-10
Fix requirements

ardunn:
The tables requirement doesn't _seem_ to be called anywhere? Having it causes some installation problems on macs without hdf5 support. I'd say that requirement can be easily deleted.

`requests` is silently required for running the documentation tests. On 99% of python installations this is fine, but on bare-bones docker installations these tests will not run because `requests` doesn't seem to be present.


ardunn:
should probably be merged at the same time as #465

shijingsun-TRI:
I have removed tables in the requirements & merged #465.

patrickherring-TRI (approved):
LGTM

PR #518, opened 2022-01-21
Allow biologic txt files to structure

shijingsun-TRI:


beep/structure/biologic.py
```diff
@@ -119,14 +119,16 @@ def from_file(cls, path, mapping_file=None):
...
-        data.loc[data.step_index % 2 == 0, "charge_capacity"] = abs(
-            data.charge_capacity
-        )
-        data.loc[data.step_index % 2 == 1, "charge_capacity"] = 0
-        data.loc[data.step_index % 2 == 1, "discharge_capacity"] = abs(
-            data.discharge_capacity
-        )
-        data.loc[data.step_index % 2 == 0, "discharge_capacity"] = 0
+        # not applicable to .txt -> check raw data formats before running the following conversion
+        if os.path.splitext(path)[1] == ".csv":
```
patrickherring-TRI, line 123: Is there a reason that the csv file will be different than the txt file?

beep/structure/tests/test_cyclerpaths.py
```diff
@@ -358,9 +358,42 @@ def test_from_csv(self):
         self.assertAlmostEqual(dp.structured_data["test_time"].max(), 101972.885, 3)
 
     def test_from_txt(self):
```
patrickherring-TRI, line 360: Let's keep the existing test and just create a new test for this bug. If some of the tests are failing we can take a look at what needs to be altered to make them pass.

patrickherring-TRI (changes requested):
Looks like you were able to resolve the issue! We should keep the existing tests so that we don't reduce functionality.

beep/structure/biologic.py
shijingsun-TRI: Now txt and csv files are treated the same. A test was added for a new txt file.

beep/structure/tests/test_cyclerpaths.py
shijingsun-TRI: Resolved. Updated the existing tests and created a new test.

shijingsun-TRI:
> Looks like you were able to resolve the issue! We should keep the existing tests so that we don't reduce functionality.

Existing tests are kept and a new test is now added.

beep/structure/biologic.py
```diff
@@ -119,14 +119,14 @@ def from_file(cls, path, mapping_file=None):
...
-        )
-        data.loc[data.step_index % 2 == 0, "discharge_capacity"] = 0
+        # data.loc[data.step_index % 2 == 0, "charge_capacity"] = abs(
+        #     data.charge_capacity
+        # )
+        # data.loc[data.step_index % 2 == 1, "charge_capacity"] = 0
+        # data.loc[data.step_index % 2 == 1, "discharge_capacity"] = abs(
+        #     data.discharge_capacity
+        # )
+        # data.loc[data.step_index % 2 == 0, "discharge_capacity"] = 0
```
ardunn, line 129: @shijingsun-TRI Is this code no longer needed even for the existing csv files?

shijingsun-TRI: That's not needed for biologic csv files we are working with now.

ardunn: Would it still support the existing biologic files or does that code truly have no use?

patrickherring-TRI (approved):
LGTM

beep/structure/biologic.py
patrickherring-TRI: I think the assumption that the charge and discharge steps will alternate does not generalize since there might be rest steps, etc. Also it looks like all of the files we get so far are resetting the charge and discharge capacity values, so this does not seem to be necessary any more.

shijingsun-TRI: I don't see the use of these lines even for the existing biologic files.

ardunn: Ok in that case I'd vote to just remove them rather than have them as comments.

beep/structure/biologic.py
```diff
@@ -119,14 +119,14 @@ def from_file(cls, path, mapping_file=None):
...
         data = pd.DataFrame(data)
-        data.loc[data.step_index % 2 == 0, "charge_capacity"] = abs(
-            data.charge_capacity
-        )
-        data.loc[data.step_index % 2 == 1, "charge_capacity"] = 0
-        data.loc[data.step_index % 2 == 1, "discharge_capacity"] = abs(
-            data.discharge_capacity
-        )
-        data.loc[data.step_index % 2 == 0, "discharge_capacity"] = 0
+        # data.loc[data.step_index % 2 == 0, "charge_capacity"] = abs(
```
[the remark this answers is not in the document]
shijingsun-TRI: Okay!

PR #523, opened 2022-01-24
Fix bug with loading processed file from a different environment

patrickherring-TRI:
There is a bug when reloading a processed file that was created with a different environment. The schema_path for validation is set as an absolute path in the first env, but when the second env tries to load the file, the same path does not exist. 

This PR adds a line where it tries to look for the schema in the default VALIDATION_SCHEMA_DIR if it cannot find the absolute path.


shijingsun-TRI (approved):
LGTM

ardunn:
Ah I see. Looks ok to me, but.... I am slightly worried about the versioning of the various schemas. The schema attr was mostly meant to just be a tool to refer back to which schema the datapath was originally validated with, **not** as a lookup tool for a schema file in different envs. In this sense, datapaths are not really portable. 


Something we could do instead is just save the entire schema as a dict **along** with the pathname. The dict would be the portable ground truth and the pathname would be a reference for where it came from originally. But iirc this would require changing some of the validation code. @patrickherring-TRI if you are looking to make datapaths portable in this way, could you add a `#todo` in the code tagging me and I will make a new issue for this? This current solution might cause problems downstream.

beep/structure/base.py
```diff
@@ -200,6 +200,8 @@ def __init__(self, raw_data, metadata, paths=None, schema=None, impute_missing=T
 
             if os.path.exists(abs_schema):
                 self.schema = abs_schema
+            elif os.path.exists(os.path.join(VALIDATION_SCHEMA_DIR, os.path.split(schema)[-1])):
+                self.schema = os.path.join(VALIDATION_SCHEMA_DIR, os.path.split(schema)[-1])
```
ardunn, line 204: may cause silent conflicts

patrickherring-TRI: Could you elaborate a little on the conflicts that you anticipate?

patrickherring-TRI:
Yeah, I would rather not have any dependence. I am just not sure how to treat the "reloading" situation differently from the situation where we are about to structure and need to make sure that the schema exists.

beep/structure/base.py
ardunn: In env1, `/path1/to/some_schema.yaml` and the file contains schema A
In env2, `/path2/to/some_schema.yaml` and the file contains schema B, just coincidentally under the same name

Perhaps not of grave concern, but just looking ahead here

Also I might have misunderstood how this is working

PR #552, opened 2022-02-24
CV capacity summary statistic

d-cogswell:
This adds a CV capacity summary statistic for both regular and diagnostic cycles, which is distinct from the CV time and current stats I added previously. I included a regression test.

After merging I would appreciate a re-run of structuring on AWS for PreDiag and PredictionDiagnostics data sets.

### DIFF
--- a/beep/structure/base.py
+++ b/beep/structure/base.py
@@ -885,6 +885,8 @@ class BEEPDatapath(abc.ABC, MSONable):
             raw_cycle = self.raw_data.loc[self.raw_data.cycle_index == cycle]
             charge = raw_cycle.loc[raw_cycle.current > 0]
             CV = get_CV_segment_from_charge(charge)
+            if CV.empty:
+                logger.debug(f"Failed to extract CV segment for cycle {cycle}!")
             CV_time.append(get_CV_time(CV))
             CV_current.append(get_CV_current(CV))
             CV_capacity.append(get_CV_capacity(CV))
@@ -1106,6 +1108,8 @@ class BEEPDatapath(abc.ABC, MSONable):
             # Charge is the very first step_index
             CCCV = raw_cycle.loc[raw_cycle.step_index == raw_cycle.step_index.min()]
             CV = get_CV_segment_from_charge(CCCV)
+            if CV.empty:
+                logger.debug(f"Failed to extract CV segment for diagnostic cycle {cycle}!")
             CV_time.append(get_CV_time(CV))
             CV_current.append(get_CV_current(CV))
             CV_capacity.append(get_CV_capacity(CV))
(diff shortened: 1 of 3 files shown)


shijingsun-TRI (approved):
LGTM, new functions will show up in the pipeline following the next release of the package.

ardunn:
tests failed due to a coveralls error, once they pass we should merge

beep/structure/base.py
```diff
@@ -1568,3 +1574,19 @@ def get_CV_current(CV):
...
+    Helper function to compute CV capacity.
+
+    Args:
+        CV (pd.DataFrame): CV segement of charge
+
+    Returns:
+        (float): charge capacity during the CV segment
+
+    """
+    if isinstance(CV, pd.DataFrame):
```
ardunn, line 1590: if there is no direct need for this `isinstance`, I'd recommend just removing it. If something gets passed into get_CV_capacity which is not a df, we would want it to throw an error. Currently it would just return `None` and might result in silent, difficult-to-debug, downstream errors

beep/structure/base.py
```diff
@@ -1568,3 +1574,19 @@ def get_CV_current(CV):
...
+        if len(CV):
```
ardunn, line 1591: same with this length check, I am not sure we'd want to return `None` here if the df is empty?

ardunn (changes requested):
I would amend the `get_CV_capacity` function to throw the correct exceptions when bad input is given; returning `None` using `isinstance` checking might cause silent errors down the line

beep/structure/base.py
d-cogswell: These checks are handling rare instances where get_CV_segment_from_charge() fails to extract a CV segment (maybe it doesn't exist, or tolerances need to be tweaked) and returns None or and empty DataFrame. When this happens, the statistic gets a value of None (which is helpful for debugging) and the data structuring just moves on to the next cycle.

Can you think of a better way to passively handle this type of failure?

ardunn: I think the question to be answered is whether the failure should be handled passively or actively. 

If it is passive, I would advocate for a minimum of an ERROR level log message.

In my experience, the most difficult types of failures to debug are those that fail silently, then require backtracking through the code to see where a certain value gets assigned. I'd generally vote for an exception, warning, or log message to be thrown

ardunn: @d-cogswell Let me know what you decide on and make any consequent changes and I'll merge in! 

Note if you decide to use a log ERROR for passive failures, you can easily import the beep one with 

`from beep import logger`

Then in an `else` block below this `if`, we'd have something like

`logger.error("No CV segment exists for <<parameters useful for debugging?>>!")`

d-cogswell: I can try moving the error checking to the CV extraction function. Should it be a warning rather than an error? Don't errors throw an exception be default?

ardunn: I don't believe so if it is within the logging framework, but I'll leave that choice (warn vs. error) up to you!

ardunn: @d-cogswell This looks ok to me now, let me know if you are ready for me to merge it

d-cogswell: Thanks @ardunn , it's ready to merge.

PR #611, opened 2022-06-05
[WIP] Novonix ingestion

shijingsun-TRI:
- Added NovonixDatapath for structuring operation, including a conversion schema, a test file, novonix.py and unit tests.


ardunn:
Hey @shijingsun-TRI 

The current problems I see are this:

- Ingestion does not currently seem to convert the capacity reading to the correct charge_capacity and discharge_capacity columns. It just gives `NaN` for both
- Other parts of the ingestion leave vestibular Temperature readings as columns and do not convert them into the existing proper formats?
- This difference in formatting is the reason why structuring does not work. For interpolation over the charge axis, for example, by default the charge axis column is charge_capacity. But if charge capacity is `NaN`, no useful interpolation will take place.


I think I can fix these and merge in, no worries

ardunn:
Also added
- `iterate_steps_in_cycle` method of BEEPDatapath for easier overriding and modifying of new datapaths with non-conforming steps index formats

ardunn:
@shijingsun-TRI I think this is about ready to be merged, just need to fix the windows tests which have something to do with unicode encoding

ardunn (approved):

PR #638, opened 2022-07-18
Bugfix and Update Novonix 

ardunn:
@shijingsun-TRI 

# Todo
- [X] Change convention for discharge capacity to increase from beginning of discharge to end of discharge cycle. Thus discharge capacity should increase as discharge steps progress, and charge capacity should increase as charge steps progress. For each cycle, discharge capacity is now calculated within the cycle as (`max_capacity_of_charge_step_of_cycle` - `discharge_capacity_of_data_point`) for each data point in the discharge step. 
- [X] Similarly change convention for discharge energy
- [X] Allow for summary file to be passed into `NovonixDatapath.from_file` and have it be easily accessible within the class as `NovonixDatapath.external_summary`. 
- [X] **Automatically extract basic metadata from the raw Novonix file; previously no metadata was recorded.** Update novonix conversion to include metadata. 
- [X] **Fix and include `step_type_number` in schema:**
    - Noxonix conversion schema now contains fields mapping the `step_type_num` to the name of the step type. this is now in the raw data, so for any given data point you can see in plain English the step type (e.g, point 999 was `cv_hold_charge`)
    - Novonix conversion schema also now maps these step types directly onto whether they are charge or discharge steps. 
- [X] Add NovonixDatapath to the docs and data requirements pages. 
- [X] New tests ensuring discharge capacity is monotonically increasing as dchg step progresses
- [X] New tests for ensuring Novonix metadata is correctly extracted
- [X] New tests for ensuring Novonix test file actually passes its own validation

More general improvements:
- [X] Better `__repr__` for looking at cycler run metadata
- [X] **Include ability to mask (exclude) cycle indices from interpolation**
    - [X] Add new test for excluding these cycle indices


ardunn:
> > # Need input
> > 
> > 1. The capacity on the discharge steps for the Novonix test file (just looking at the raw file, not doing **any beep processing**) seems to be increasing... is this actually correct? Shouldn't the raw value of capacity (before doing beep processing) be decreasing as the discharge steps progress?
> > 2. Should we apply the same logic to `discharge_energy` (i.e., increasing as dchg step progresses)?
> > 3. Should the externally loaded summary file be merged with the beep summary, or kept as a separate attribute such as `NovonixDatapath.external_summary`? I am leaning towards the latter because it is less code and less potential for random errors (such as beep detecting a different number of cycles than the Novonix summary file...)
> 
> 1. This is incorrect. The error arises due to Step Number and Step Type are not 1-to-1 match in Novonix raw file. Let's use Step Type to identify charge vs. discharge steps in raw data.
> 2. Yes, we should apply the same logic
> 3. Yes, I also agree having an optional NovonixDatapath.external_summary is better than merging the internal and external summaries, this way we can avoid confusions from the users on data source.

For (2) and (3) this sounds good. 

For 1, I am not sure how this applies. We are already using only `step_type` (as far as I can tell) to determine charge and discharge steps. I mean if you look at the raw file (like load it in a pandas dataframe directly, or just look at the capacity numbers in the raw file), you'll see the capacity is monotonically increasing regardless of charge/discharge step or step type. 


To test this, try cutting out everything above the header line in the raw file and name it `tmp.csv`. Then run:

```
import pandas as pd

raw = pd.read_csv("tmp.csv", sep='\t', header=None)
data = raw
raw = raw[0].str.split(',', expand=True)
headers = raw.iloc[0]
data = pd.DataFrame(data.values[1:], columns=headers, index=None)

print(raw)

print(raw["Capacity (Ah)"].is_monotonic)
```

It outputs 

```
0              Date and Time Cycle Number  ... Step Number Step position
0     2022-04-11 10:58:07 AM            1  ...           1             0
1     2022-04-11 10:58:08 AM            1  ...           1             0
2     2022-04-11 10:58:09 AM            1  ...           1             0
3     2022-04-11 10:58:10 AM            1  ...           1             0
4     2022-04-11 10:58:11 AM            1  ...           1             0
...                      ...          ...  ...         ...           ...
3937   2022-04-12 4:23:59 AM            1  ...           3             0
3938   2022-04-12 4:24:02 AM            1  ...           3             0
3939   2022-04-12 4:24:05 AM            1  ...           3             0
3940   2022-04-12 4:24:08 AM            1  ...           3             0
3941   2022-04-12 4:24:11 AM            1  ...           3             0

[3942 rows x 15 columns]
True
```

So the capacity is monotonically increasing throughout **all** the different steps... what is going on here?!?

shijingsun-TRI:
> So the capacity is monotonically increasing throughout **all** the different steps... what is going on here?!?

Thanks a lot for sharing the test! Super helpful for us to debug. I think I figured out why - the test file we are using is a shortened version of the full cycling, because of the high complexity of this protocol, the shortened version didn't capture all the steps.

ardunn:
@shijingsun-TRI I think this is ready for review now

shijingsun-TRI (approved):
LGTM

PR #659, opened 2022-08-16
Bugfix novonix ingestion for cycles with no charge step

ardunn:
- Previously, novonix ingestion failed silently (nans for all values of dchg cap and energy) on cycles where there are no charge steps, as the caps/engs could not be converted to montonically increasing since there was no max value from a charge step
- Now, explicitly accounts for these cycles with no charge step by setting the maximums from the highest raw `capacity` value within the cycle. Now the convention for these cycles is `discharge_capacity` = `max_discharge_capacity_within_this_cycle` - `capacity` (and similarly for energy). The other cycles are unchanged
- Also includes tests specifically for this case now in `test_cyclerpaths.TestNovonixDatapath.test_file_long` by looking explicitly at cycle 4.
- Also includes a warning when this extra conversion is needed. It looks like this:
    - `2022-08-16 13:13:05 WARNING  No charge steps found in cycles [4]! Using highest capacity reading to determine convention.`
- Also fixes a bug where interpolation axis was not shown if axis != `voltage`

For reference, cycle 4 of the long test file now looks like this once ingested:

image

And slinear interpolated (resolution=100)/structured

image

### DIFF
--- a/beep/structure/base.py
+++ b/beep/structure/base.py
@@ -612,10 +612,12 @@ class BEEPDatapath(abc.ABC, MSONable):
         if not exclude_cycles:
             exclude_cycles = []
 
-        if not desc:
+        if not desc and axis == "voltage":
             desc = \
                 f"Interpolating {step_type} ({v_range[0]} - {v_range[1]})V " \
                 f"({resolution} points)"
+        elif not desc:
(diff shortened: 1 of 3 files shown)


shijingsun-TRI (approved):
Thanks for implementing the change, LGTM.

PR #664, opened 2022-08-19
Fix docs

ardunn:
Docs were somehow corrupted


ardunn:
@shijingsun-TRI sort of off topic from this PR, but do you know why the coverage is failing for every PR? Something must be misconfigured, because it seems like no matter what the change is, the coverage somehow decreases...

shijingsun-TRI (approved):

shijingsun-TRI:
@ardunn Indeed, I also noticed it recently. From coveralls, it looks s3 and processing diagnostics data are two places with most uncovered code.

PR #687, opened 2022-09-27
Update NovonixDatapath with improved parsing capability

ardunn:
Update NovonixDatapath to handle new formats from new cycler firmware/software

Also includes tests

This should maintain backwards compatibility with the older Novonix software as well


shijingsun-TRI (approved):
Thanks Alex, LGTM

PR #794, opened 2023-05-16
[WIP] update scipy version

JosephMontoya-TRI:
### Changes proposed in this pull request:
- [WIP] for updating scipy version to inspect test differences.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # Base requirements
 numpy==1.22.2
 monty[yaml]==2022.4.26
-scipy==1.5.4
+scipy==1.10.1
 scikit-learn==1.0.2
 pandas==1.4.1
 tqdm==4.62.3


ardunn:
@JosephMontoya-TRI so according to this: https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html

I think the newest scipy changes what slinear does, though I'm not exactly sure how it differs from linear. 

Just changing the interpolation keyword to be "linear" removes most of the errors for 'duplicate x', and the remaining seem to be problems with the precision of the interpolation like this:


```
for col_name in y_at_point.columns:
>           self.assertAlmostEqual(
                pred[col_name].iloc[0], y_at_point[col_name].iloc[0],
                places=2
            )
E           AssertionError: 3320.0375418686945 != 3319.9971375 within 2 places (0.040404368694453296 difference)
```

JosephMontoya-TRI:
Yeah I came to a similar conclusion the last time I tried this (which was like 3 years ago). I think we should probably just update the tests, change the interpolation keyword, and cross our fingers that this isn't breaking too many things downstream (if it's too problematic we can always change it back). @shijingsun-TRI @pkherring does that sound okay?

pkherring:
Yeah, I agree. Just update the test values and upgrade scipy

